Issue #132, opened 2014-02-05
Rename Zip module to RubyZip


weshatheleopard:
Gems that include <code>rubyzip</code> conflict with gems that include <code>zip</code> since both call their main module <code>Zip</code>. A way to separate the two is absolutely necessary if we want to avoid dirty hackery to the tune of:

<pre>
unless ::Object.const_defined?(:RubyZip, false)
  if ::Object.const_defined?(:Zip, false) then
    if ::Zip.const_defined?(:File, false) then
       puts "DEBUG: RubyZip detected"
      ::RubyZip = ::Zip
    else
       puts "DEBUG: Conflicting Zip detected"
      zip_backup = ::Zip
      ::Zip = nil
      require 'rubygems'
      gem 'rubyzip', :require => 'zip'
      ::RubyZip = ::Zip
      ::Zip = zip_backup
    end
  else
     puts "DEBUG: Zip not detected at all"
    require 'rubygems'
    gem 'rubyzip', :require => 'zip'
    ::RubyZip = ::Zip
  end
end
</pre>

Note that the main include file should also be renamed to <code>rubyzip.rb</code>, otherwise loading the right gem also turns into a Russian roulette game.

mikeLspohn:
Anything on this? This is causing me issues as well. Sure, zipruby may be unmaintained as you say, but that doesn't mean existing apps aren't using it, or that there aren't other gems depending on it that people may be using.

sandstrom:
Much better to ask the owner of the unmaintained `zip` gem if they're willing to release the name.

Gems are still versioned, so as long as the major version is switched it would be no different than other breaking changes (in other words, replacing the entire underlying library for a gem is not different from other types of breaking changes).

jspanjers:
I think the top level class should be Rubyzip (not RubyZip).

VSPPedro:
+1

I have a Zip model. When executing Rails.application.eager_load! I got this error: 

```
TypeError: Zip is not a class
/usr/local/bundle/ruby/2.7.0/gems/rubyzip-2.3.2/lib/zip/constants.rb:1: previous definition of Zip was here
```

What are your thoughts on this change? If there's anything I can do, I'd love to help.

jcw-:
This gem is missing a `lib/rubyzip.rb` file, preventing `Bundler.require(:default, Rails.env)` in a Rails application from loading this gem without an additional explicit require of `zip.rb`.

To work around this, you can declare the gem like this:
```
[...]

PR #145, opened 2014-03-20
Renaming the Zip module to RubyZip

davidhooey:
- To cover Issue https://github.com/rubyzip/rubyzip/issues/132
- All tests pass.
- Successfully used the renamed module to avoid conflict with the rawr module. 
- Updated README.md but may need more clarification from the owners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,25 +5,22 @@
 
 rubyzip is a ruby library for reading and writing zip files.
 
-## Important note
-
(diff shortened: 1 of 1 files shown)


sandstrom:
I agree aligning the module name and the gem-name would be nice. But I'd rather prefer to have the gem renamed to `zip`

PR #179, opened 2014-08-06
Add support for reading ZIP files utilising AES encryption

muz:
Working AES-128, -192, -256 decryption

Caveats:
- Doesn't currently check the authentication code
- Won't read zip files which have a data description block without
  the signature. (See notes in code)
- Has no tests yet

### DIFF
--- a/lib/zip/input_stream.rb
+++ b/lib/zip/input_stream.rb
@@ -129,15 +129,16 @@ module Zip
       @current_entry
     end
 
-    def get_decompressor
-      case
-      when @current_entry.nil?
+    def get_decompressor(compression_method = nil)
+      compression_method ||= @current_entry.compression_method unless @current_entry.nil?
+      case compression_method
+      when nil
         ::Zip::NullDecompressor
-      when @current_entry.compression_method == ::Zip::Entry::STORED
+      when ::Zip::Entry::STORED
         ::Zip::PassThruDecompressor.new(@archive_io, @current_entry.size)
-      when @current_entry.compression_method == ::Zip::Entry::DEFLATED
+      when ::Zip::Entry::DEFLATED
         ::Zip::Inflater.new(@archive_io)
-      when @current_entry.compression_method == ::Zip::Entry::ENCRYPTED
+      when ::Zip::Entry::ENCRYPTED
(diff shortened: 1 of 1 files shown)


johnnyshields:
@muz @jphastings my team is working on traditional encryption (work-in-progress) here: johnnyshields/rubyzip. We've come up with a structure that I believe is a bit more extensible--should be easy to plug AES into it. Please take a look.

jphastings:
Will do @johnnyshields, where would you recommend we start looking - in the master branch?

johnnyshields:
@jphastings here's a diff https://github.com/johnnyshields/rubyzip/compare/rubyzip:master...master (Github is so cool!)

We've added the structures for both encryption and decryption. I think one of the biggest things to determine is what the public interface should be, in particular `OutputStream#open` and `#write_buffer` methods and how we can support both aes and traditional encryption on them.

The code itself is a work in progress, doesn't seem to be 100% working yet but the structure is there. Copying @matsu911 to the thread.

johnnyshields:
FYI I've raised the PR for Traditional Encryption

simonoff:
@muz can you rebase with current master?

cielavenir:
I have skimmed the code and noticed that AES encryption is compression method 99. So actually AES has to be implemented as a decompressor, which makes current "transparent decryption" difficult.

bensomers:
No action on this PR for several years now - what are thoughts like on the prospects for this PR/this feature more generally?

jphastings:
Hey @bensomers; I used to work with @muz while the company we completed this work for still existed. I can put some work into getting this up to par if @cielavenir's concerns aren't problematic to the maintainers.

bensomers:
I'd be a fan of that, though it's really the maintainers we'd want to hear from before you spend the effort.

I actually need to be writing out AES files, not just reading them, so this would only be a start point for my needs - but it's pretty silly for me to submit a PR for writing before the capability to read is in.

tramuntanal:
Hello everybody, is there a plan to add AES support (read/write) to rubyzip? It may be interesting for me to work on this with a small guidance.

johnnyshields:
Hi, I won't be working on this. With some googling you can probably find a ruby implementation or a python/javascript implementation that could be ported.

PR #183, opened 2014-08-21
Fixed directories usage example

novito:
The example was not working as it was written. Now it should work.


jdleesmiller:
Thanks (belatedly) for your help with updating the README, but it looks like the example in the README was substantially changed in #297 (and probably before then, too), so these changes no longer apply.

PR #213, opened 2015-01-26
Force UTF-8 encoding in file names

SeTeM:
This PR is pretty similar to the rubyzip/rubyzip/pull/191, but it has test and uses `String#force_encoding`

```
-> % unzip -t Безымянныи<0306>-1-1.xhtml.zip
Archive:  Безымянный-1-1.xhtml.zip
    testing: Безымянный-1-1.xhtml   OK
    testing: __MACOSX/                OK
    testing: __MACOSX/._Безымянный-1-1.xhtml   OK
No errors detected in compressed data of Безымянный-1-1.xhtml.zip.
```


simonoff:
@SeTeM what the reason of this code? If you want to set names in UTF-8 why not doing it before put it to zip file?

SeTeM:
Archives are not mine. I need to extract it's files and process them.

simonoff:
@SeTeM So what the problem to doing extraction and inside this extraction doing forcing file names to UTF-8? It's only your specific case.

PR #222, opened 2015-02-19
Case insensitivity option for #find_entry

rastarobot:
So here's another take on https://github.com/rubyzip/rubyzip/pull/217
In the end I decided that once case-insensitivity is being implemented in this single method, adding an extra configuration option would not be justified, as that option would not turn `rubyzip` case-insensitive throughout.
Pull if you consider this solution acceptable.

### DIFF
--- a/lib/zip/entry.rb
+++ b/lib/zip/entry.rb
@@ -221,7 +221,7 @@ module Zip
     def read_local_entry(io) #:nodoc:all
       @local_header_offset = io.tell
 
-      static_sized_fields_buf = io.read(::Zip::LOCAL_ENTRY_STATIC_HEADER_LENGTH)
+      static_sized_fields_buf = io.read(::Zip::LOCAL_ENTRY_STATIC_HEADER_LENGTH) || ''
 
       unless static_sized_fields_buf.bytesize == ::Zip::LOCAL_ENTRY_STATIC_HEADER_LENGTH
         raise Error, "Premature end of file. Not enough data for zip entry local header"
--- a/lib/zip/entry_set.rb
+++ b/lib/zip/entry_set.rb
@@ -13,8 +13,10 @@ module Zip
       @entry_set.include?(to_key(entry))
     end
 
-    def find_entry(entry)
-      @entry_set[to_key(entry)]
+    def find_entry(entry, case_sensitively = true)
+      return @entry_set[to_key(entry)] if case_sensitively
+      entry = @entry_set.find { |k, _| k.downcase == to_key(entry).downcase }
+      entry.last if entry
     end
 
     def <<(entry)
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ Gemfile.lock
 +samples/*.zip.*
 coverage
 pkg/
+.ruby-gemset
+.ruby-version
--- a/test/entry_set_test.rb
+++ b/test/entry_set_test.rb
@@ -65,6 +65,12 @@ class ZipEntrySetTest < MiniTest::Test
     assert_equal(ZIP_ENTRIES, @zipEntrySet.entries)
   end
 
+  def test_find_entry
+    # by default, #find_entry is case-sensitive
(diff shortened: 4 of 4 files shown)


dmitry:
Looks good to me. :+1:

lib/zip/entry.rb
```diff
@@ -221,7 +221,7 @@ def unpack_local_entry(buf)
     def read_local_entry(io) #:nodoc:all
       @local_header_offset = io.tell
 
-      static_sized_fields_buf = io.read(::Zip::LOCAL_ENTRY_STATIC_HEADER_LENGTH)
+      static_sized_fields_buf = io.read(::Zip::LOCAL_ENTRY_STATIC_HEADER_LENGTH) || ''
```
simonoff: What the reason of it?

rastarobot: It fixes the tests. They were failing on JRuby on this line as it returned `nil`, and thus subsequent `static_sized_fields_buf.bytesize` call resulted in `NoMethodError`

simonoff:
You will not add setting for this flag?

simonoff:
Ok

lib/zip/entry_set.rb
```diff
@@ -13,8 +13,10 @@ def include?(entry)
       @entry_set.include?(to_key(entry))
     end
 
-    def find_entry(entry)
-      @entry_set[to_key(entry)]
+    def find_entry(entry, case_sensitively = true)
+      return @entry_set[to_key(entry)] if case_sensitively
+      entry = @entry_set.find { |k, _| k.downcase == to_key(entry).downcase }
```
weshatheleopard: Guys, come on, this is suboptimal.

What should be done instead:
`
def to_key
   result = (blah blah ...)
   result = result.downcase! if options.case_insensitive_search
   result
end
`

This way, the entire code will operate properly in case if `options.case_insensitive_search` is `true`.

rastarobot: Not really. Say we're given this:

```ruby
@entry_set = {
  'FileNameOne' => { ... },
  'FileNameTwo' => { ... }
}
```

and are trying to find the entry by calling

```ruby
find_entry('FILENAMEONE')
```

..having set ```options.case_insensitive_search``` to ```false```. It doesn't take a scientist to see that the entry will not be retrived:

```ruby
to_key('FILENAMEONE')
=> 'filenameone'
@find_entry['filenameone']
=> nil
```

because it's sort of clear that hash keys need to be downcased, too. Or isn't it? Try this in ```irb```:

```ruby
a = {'ABC' => 1, 'Abc' => 2};
a['AbC']
```

weshatheleopard: @rastarobot, you are 100% correct that "hash keys need to be downcased too", but aren't the hash entry keys formed by calling `to_key` already?

Something like 
```
def add_file_to_hash(name, entry)
  @hash[to_key(name)] = entry
   ...
end
```
That's the entire point why I'm telling you to modify `to_key`, not `find`.

weshatheleopard: @rastarobot, in short. *this* is how you do it: https://github.com/rubyzip/rubyzip/compare/master...weshatheleopard:master (sorry, I couldn't run tests because `simplecov` refuses to install on my home box, and I have no time to figure out why).

rastarobot: Good job. Turns out I had misconception of the inner workings of the gem.
And just for the record. It should be added somewhere to the documentation that similarly looking but differently cased entries may be lost on opening an archive, as EntrySet keeps them in a hash using filenames for hash keys. I.e. if the archive includes File.txt, fiLe.txt and filE.txt only one of them will make its way to the resulting rubyzip object.

weshatheleopard: OK, I was able to run tests. Everything works as expected. "Опыт не пропьёшь" ©☺

https://github.com/rubyzip/rubyzip/pull/224

PR #226, opened 2015-03-20
Add configs of rubocop gem and fix some rubocop warnings

ShockwaveNN:
I run rubocop (https://github.com/bbatsov/rubocop) gem on this project and fix some of problem to try it out.
If maintainers of this project think this is good idea - I can continue to fix code style warnings and creating pull requests for it


simonoff:
@ShockwaveNN If you have a time then ok. But rubocop is not enough. You can look on reek and cane.

.rubocop.yml
```diff
@@ -0,0 +1,2 @@
+inherit_from:
+    - .rubocop_todo.yml
```
simonoff: Please rename file to .rubocop_rubyzip.yml

ShockwaveNN: Are you sure about it?
Without renaming I can just run `rubocop` command from root of project and it will use this config file, but if i rename - I should run `rubocop -c .rubocop_rubyzip.yml`

simonoff: Why? you renaming file and renaming entry in `.rubocop.yml`

ShockwaveNN: Oh, I get it. I thought you mention `.rubocop.yml` file, but you mention `.rubocop_todo.yml` file

ShockwaveNN:
I fix some trivial problems, I stop now, but try to continue tomorrow

lib/zip/entry.rb
```diff
@@ -83,7 +83,7 @@ def time
       end
     end
 
-    alias :mtime :time
+    alias_method :mtime, :time
```
simonoff: For what reason you did it?

simonoff: There is no sense to use alias_method because we don't need chains. So better use alias instead of alias_method

lib/zip/entry_set.rb
```diff
@@ -21,13 +21,13 @@ def <<(entry)
       @entry_set[to_key(entry)] = entry
     end
 
-    alias :push :<<
+    alias_method :push, :<<
```
simonoff: There is no sense to use alias_method because we don't need chains. So better use alias instead of alias_method

lib/zip/entry_set.rb
```diff
@@ -21,13 +21,13 @@ def <<(entry)
...
 
     def size
       @entry_set.size
     end
 
-    alias :length :size
+    alias_method :length, :size
```
simonoff: There is no sense to use alias_method because we don't need chains. So better use alias instead of alias_method

lib/zip/entry_set.rb
```diff
@@ -49,7 +49,7 @@ def entries
 
     # deep clone
     def dup
-      EntrySet.new(@entry_set.map { |key, value| value.dup })
+      EntrySet.new(@entry_set.map { |_key, value| value.dup })
```
simonoff: This can be transformed to `values.map(&:dup)`

lib/zip/extra_field.rb
```diff
@@ -3,7 +3,7 @@ class ExtraField < Hash
     ID_MAP = {}
 
     def initialize(binstr = nil)
-      binstr and merge(binstr)
+      binstr && merge(binstr)
```
simonoff: No in this code need exactly `and`

lib/zip/extra_field.rb
```diff
@@ -69,7 +69,7 @@ def to_local_bin
       ordered_values.map! { |v| v.to_local_bin.force_encoding('BINARY') }.join
     end
 
-    alias :to_s :to_local_bin
+    alias_method :to_s, :to_local_bin
```
simonoff: There is no sense to use alias_method because we don't need chains. So better use alias instead of alias_method

lib/zip/extra_field.rb
```diff
@@ -83,8 +83,8 @@ def local_size
       to_local_bin.bytesize
     end
 
-    alias :length :local_size
-    alias :size :local_size
+    alias_method :length, :local_size
+    alias_method :size, :local_size
```
simonoff: There is no sense to use alias_method because we don't need chains. So better use alias instead of alias_method

lib/zip/entry.rb
ShockwaveNN: Yeah, agree with you based on https://github.com/bbatsov/ruby-style-guide#alias-method, I redone it soon in all cases

lib/zip/extra_field.rb
ShockwaveNN: If this code need `and` to implement control flow - I think it need to be redone without usage of `and`, because of https://github.com/bbatsov/ruby-style-guide#no-and-or-or

Also since this code needed exactlye 'and' I think there is need of test, that failed on change on `&&` since not a single test failed after all my commits

lib/zip/entry_set.rb
ShockwaveNN: Like this ` EntrySet.new(@entry_set.map(&:dup))`?
Not working, a lot of test failed

simonoff: Like `EntrySet.new(@entry_set.values.map(&:dup))`

lib/zip/extra_field.rb
ShockwaveNN: @simonoff Can you help me to simplify this statement without usage of 'and', I read http://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby, but still not quite sure about it

marcuserronius: Hi, I've been lurking here for a while, but thought I could answer this
question...

"&&" is usually used in its traditional sense, as a logical operator.
However, in other languages, it has some history of use as flow; being used
as a shorthand "if...then", or to replace the ternary operator. Ruby was
designed with the "and" keyword for those cases.

If this section was to be rewritten, a better way than using "&&" would be:

    merge(binstr) if binstr

The fact that it is more understandable rewritten as a conditional is the
biggest indication that the choice of "and" is better than "&&".

Hope this helps!

Mark

ShockwaveNN: @marcuserronius thank you

ShockwaveNN:
I think I fixed almost all trivial cases, so I think it may be merged with master, because after that point fixes will be more dramatical, and may be lost in existing changes. Existing changes is small, but there is too much number of them. Also future fixes volumes will not be so high, as now

simonoff:
Can you rebase with latest master?

ShockwaveNN:
@simonoff Yes, of course I can rebase. I try to do it tomorrow

ShockwaveNN:
@simonoff Done rebasing, also update rubocop and fix some more problems

PR #237, opened 2015-06-24
Do something more expected with new file permissions.

### LINKED ISSUE #204: 0600 ZIP file permissions with umask 0022
I do:

``` ruby
Zip::File.open(zip_file, Zip::File::CREATE) do |zip|
  begin
    zip.add(name, bin)
  rescue Zip::ZipEntryExistsError
    zip.replace(name, bin)
  end
end
```

Is there a reason, I get `0600` file permissions on my `zip_file` instead of `0644`, I would expect? Diagnostics:

``` sh
$ umask
0022
$ ruby --version 
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-darwin12.0]
```

I googled that and found out an advice to use `File.chmod` implicitly. Am I overlooking something? In particular:
1. Why rubyzip does not default to `umask`-based permissions? Is it documented?
2. Are there any handles to get `umask`-based behaviour, not a hardcoded `0600`? Didn't find them in docs.

hainesr:
This fixes #204.

Instead of inheriting the permissions from the tmp directory, new files should
have permissions that reflect the defaults on the system taking umask into
account.

It seems that (unix) permissions of 666 - umask are as close to a standard as
anything [1] and that 'touch' uses this. On Windows it seems sensible to just
use 644 directly [2].

[1] http://unix.stackexchange.com/a/102080
[2] http://ruby-doc.org/core-1.9.3/File.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,16 @@ fruit/orange
 
 After this, entries in the zip archive will be saved in ordered state.
 
+### Default permissions of zip archives
+
+On Posix file systems the default file permissions applied to a new archive
+are (0666 - umask), which mimics the behavior of standard tools such as `touch`.
+
+On Windows the default file permissions are set to 0644 as suggested by the
+[Ruby File documentation](http://ruby-doc.org/core-2.2.2/File.html).
+
+When modifying a zip archive the file permissions of the archive are preserved.
+
 ### Reading a Zip file
 
 ```ruby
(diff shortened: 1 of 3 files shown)


gregawoods:
Nice! Pulled the code down and verified that it works as expected in my project. 👍

simonoff:
@hainesr great! Would be awesome if you will update README about this behavior. Thank you!

hainesr:
No problem, done!

simonoff:
@hainesr thanks!

chillum:
guys, will we have a new release with that?

in 1.1.7 that patch is absent and there is no release after that (nearly a year ago).

PR #238, opened 2015-07-02
Accept StringIO in Zip.open_buffer

pwnall:
It'd be really handy if `open_buffer` accepted a StringIO parameter, because I could just give it the `body` attribute of a Rails request, which can either be an IO or a StringIO.

The `open_buffer` implementation wraps String buffers into StringIO instances, so it appears that it should be able to use a StringIO without any deep changes. I updated the checks and added a test verifying the new code.

Will you please consider merging this?

Thank you very much for rubyzip!

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -115,8 +115,8 @@ module Zip
       # (This can be used to extract data from a
       # downloaded zip archive without first saving it to disk.)
       def open_buffer(io, options = {})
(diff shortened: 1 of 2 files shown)


pwnall:
That was amazingly quick! Thank you! :+1:

programble:
This fix would be great to have. When can we expect a new release to include it?

Issue #233, opened 2015-05-10
NoMemoryError when zipping file that does not fit in RAM with swap off


alexis779:
Hi,

`Zip::OutputStream` should be able to stream a file which size is bigger than the available RAM directly to the zip output file, without incurring into swapping hell or getting killed by Ruby because the process is filling up the memory without regularly flushing to disk.

Here is a `test/memory_test.rb` unit test currently failing:

```ruby
# encoding: utf-8
require 'test_helper'

class MemoryTest < MiniTest::Test

  # Make sure we can add a file that does not fit in RAM to the zip
  def test_memory
    buffer = " " * 1024
    count = 10 ** 8
    # try to write a 100 GB file
    Zip::OutputStream.open("test/data/generated/memorytest.zip") {  |zos|
      zos.put_next_entry("bigfile", nil, nil, Zip::Entry::DEFLATED, 0)

      count.times {
        zos.write buffer
      }
    }
  end

end
```

Make sure you disable swapping:
```
sudo swapoff -a
```

The test fails with:

```
bundle exec rake test TEST=test/memory_test.rb
/home/alexis/job/rubyzip/lib/zip/deflater.rb:18:in `write': failed to allocate memory (NoMemoryError)
        from /home/alexis/job/rubyzip/lib/zip/deflater.rb:18:in `<<'
        from /home/alexis/job/rubyzip/lib/zip/deflater.rb:18:in `<<'
        from /home/alexis/job/rubyzip/lib/zip/output_stream.rb:181:in `<<'
        from /home/alexis/job/rubyzip/lib/zip/ioextras/abstract_output_stream.rb:9:in `write'
        from /home/alexis/job/rubyzip/test/memory_test.rb:16:in `block (2 levels) in test_memory'
        from /home/alexis/job/rubyzip/test/memory_test.rb:15:in `times'
        from /home/alexis/job/rubyzip/test/memory_test.rb:15:in `block in test_memory'
        from /home/alexis/job/rubyzip/lib/zip/output_stream.rb:53:in `open'
        from /home/alexis/job/rubyzip/test/memory_test.rb:12:in `test_memory'
        from /var/lib/gems/2.1.0/gems/minitest-5.6.1/lib/minitest/test.rb:108:in `block (3 levels) in run'
        from /var/lib/gems/2.1.0/gems/minitest-5.6.1/lib/minitest/test.rb:205:in `capture_exceptions'
[...]

PR #250, opened 2015-10-17
Fix high memory usage due to Deflater buffering

felixbuenemann:
When support for ZipCrypto was added, an internal StringIO buffer was added to Deflater, in order to fix a decryption bug. While this worked, it caused unlimited memory growth when compressing large files.

The proper fix is to writer the encryption header in init_next_entry instead of finalize_current_entry, so the headers are written before any encrypted data. Because of this fix we can remove the buffering in
Deflater, which keeps memory usage low and allows to stream compressed data while it is written.

This should fix issue #233.


felixbuenemann:
Thanks, Alexander!

PR #252, opened 2015-10-28
Don't enable JRuby objectspace

jjb:
From the JRuby wiki:

> ObjectSpace has been disabled by default since version 1.1b1. Some users reenable ObjectSpace, using the -X+O flag (previously +O), without realizing what a massive performance hit this is. If you are running with ObjectSpace enabled, do not expect any sort of performance.

https://github.com/jruby/jruby/wiki/PerformanceTuning#dont-enable-objectspace

This was enabled in 04029115. @simonoff, do you remember the rationale? I found no discussion in any issues or PRs: https://github.com/rubyzip/rubyzip/issues?utf8=%E2%9C%93&q=objectspace

Some basic tests in my app show that nothing breaks without this enabled.

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -405,14 +405,11 @@ module Zip
     def on_success_replace
       tmpfile      = get_tempfile
       tmp_filename = tmpfile.path
-      ObjectSpace.undefine_finalizer(tmpfile)
       tmpfile.close
       if yield tmp_filename
         ::File.rename(tmp_filename, name)
         ::File.chmod(@file_permissions, name) if defined?(@file_permissions)
       end
-    ensure
-      tmpfile.unlink if tmpfile
     end
 
     def get_tempfile
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -405,11 +405,11 @@ module Zip
     def on_success_replace
       tmpfile      = get_tempfile
       tmp_filename = tmpfile.path
-      tmpfile.close
       if yield tmp_filename
         ::File.rename(tmp_filename, name)
         ::File.chmod(@file_permissions, name) if defined?(@file_permissions)
       end
+      tmpfile.close
     end
 
     def get_tempfile
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -406,7 +406,9 @@ module Zip
       tmpfile      = get_tempfile
       tmp_filename = tmpfile.path
       if yield tmp_filename
-        ::File.rename(tmp_filename, name)
+        new_file = ::File.new(name, "w")
+        IO.copy_stream tmpfile, new_file
+        new_file.close
         ::File.chmod(@file_permissions, name) if defined?(@file_permissions)
       end
       tmpfile.close
--- a/lib/zip.rb
+++ b/lib/zip.rb
@@ -30,10 +30,6 @@ require 'zip/streamable_stream'
 require 'zip/streamable_directory'
 require 'zip/constants'
 require 'zip/errors'
-if defined? JRUBY_VERSION
-  require 'jruby'
-  JRuby.objectspace = true
-end
 
 module Zip
   extend self
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -6,6 +6,11 @@ require 'digest/sha1'
 require 'zip'
 require 'gentestfiles'
 
+if defined? JRUBY_VERSION
+  require 'jruby'
+  JRuby.objectspace = true
+end
+
 TestFiles.create_test_files
 TestZipFile.create_test_zips
 


chrisseaton:
Looks like it might just be used in testing:

https://github.com/rubyzip/rubyzip/blob/master/test/test_helper.rb#L206-L207

simonoff:
@chrisseaton can you enable it only for testing? I'm not so familiar with jRuby.

chrisseaton:
Move

```ruby
if defined? JRUBY_VERSION
  require 'jruby'
  JRuby.objectspace = true
end
```

From `lib/zip.rb` to `test/test_helper.rb` I guess.

jjb:
Okay but, why is it needed for testing? Does that suggest the test environment has something very different from a typical runtime environment?

jjb:
@chrisseaton 

> it only needs to be turned on for some features

which features?

chrisseaton:
I think `id2ref` and `each_object`, unless you pass `Class`. But if your app works without turning it on then it's fine - you get an exception if you try to use a feature that is turned off.

@headius may be able to get something more definitive.

jjb:
Looks like the purpose of this line:

```ruby
ObjectSpace.undefine_finalizer(tmpfile)
```

introduced in https://github.com/rubyzip/rubyzip/pull/185 / 3b639b6c is to "prevent tempfile path from being unlinked by garbage collection".

The PR is talking about garbage collection, but the line immediately after that one is `tmpfile.close`, which of course is expected to unlink the file. Code that will persist the availability of the file after that can't be right, IMO...

I see further down that the Tempfile is renamed. I think this can be considered a misuse of Tempfile. Instead, a new File should be created, with the contents of the Tempfile streamed into it.

I'll poke around with cleaning this up and possible come up wtih a PR. Meanwhile I welcome anyone's thoughts.

/cc @maneframe @simonoff @andreibondarev @chrisseaton

headius:
> I think id2ref and each_object, unless you pass Class.

That's correct. `each_object` with a Class or Module always works. You only need to enable ObjectSpace if you need to pass `each_object` arbitrary types or use `_id2ref` to retrieve non-Fixnum objects.

The other features of ObjectSpace, like finalizers, work all the time.

jjb:
Okay, I've made #253

jjb:
@chrisseaton I see no uses of `each_object` in the codebase. Is that correct?

`_id2ref ` is only used in the test code. I'm going to tinker with replacing the use of `assert_forwarded`. Stay tuned...

jjb:
Okay I've now rebased this PR to be on top of #253, and moved the objectspace activation code to test_helper.rb as @chrisseaton suggested. See comments in d0c5db9719cd140835138477b37138d0575bf957

jjb:
> _id2ref is only used in the test code. I'm going to tinker with replacing the use of assert_forwarded. Stay tuned...

I gave up on this... will wait to hear feedback on the various other things. If the maintainers are okay with only turning on objectspace during tests, I am too.

headius:
Status? We just got a bug report from a user that had reduced performance just because they used rubyzip. We'd *REALLY* like to get this merged and released.

jjb:
> Status?

First #253 needs to be discussed and merged.

headius:
@jjb I don't see anything in #253 that needs JRuby's extended ObjectSpace features to be enabled.

mnaberez:
> First #253 needs to be discussed and merged.

Please do not merge #253 unless it is rewritten to fix the performance and data corruption issues it would introduce ([explanation](https://github.com/rubyzip/rubyzip/pull/253#discussion_r43533967)).

jjb:
@mnaberez I have an option question for you in a subsequent comment: "So, we could remove 1ad5814 from my PR and leave the other two."

lib/zip/file.rb
```diff
@@ -405,14 +405,13 @@ def check_file(path)
     def on_success_replace
       tmpfile      = get_tempfile
       tmp_filename = tmpfile.path
-      ObjectSpace.undefine_finalizer(tmpfile)
-      tmpfile.close
       if yield tmp_filename
-        ::File.rename(tmp_filename, name)
+        new_file = ::File.new(name, "w")
+        IO.copy_stream tmpfile, new_file
```
mnaberez: This pull request, #252, introduces the same [performance and data corruption issues](https://github.com/rubyzip/rubyzip/pull/253#discussion_r43533967) as #253.

jjb:
@mnaberez I'll restate the situation:

This pull request, 252, is a branch that is based on the branch in 253.

253 should first be discussed and possibly merged. Then 252 can be rebased on master and discussed and possibly merged.

Thanks!

mnaberez:
Rubyzip uses this `on_success_replace` method to build the full contents of the ZIP in a `Tempfile`.  The way the code works is that if building the ZIP was successful, the ZIP is moved into its final location by  `File.rename` from the tempfile path to the final path.

A feature of `Tempfile` on MRI that can't be disabled is that it always defines a finalizer.  The finalizer is a `Tempfile::Remover` that unlinks the file.  It looks like the way that the Rubyzip code was supposed to work is that if building the ZIP succeeded, the tempfile gets moved.  When the finalizer runs, the original path of the tempfile no longer exists, so the finalizer silently fails.

The use of `ObjectSpace` was introduced in RubyZip in 3b639b6c622ecdd21f33907e735a62bed12dcf4b (pull request #185).  Apparently there is some scenario where the tempfile is unexpectedly unlinked by the finalizer.  Since the `Tempfile` finalizer is not optional, that pull request's solution was to use `ObjectSpace.undefine_finalizer` after the fact, then delete the tempfile manually in an ensure block.  Similarly, #252 and #253 are different attempts to work around the `Tempfile` finalizer.

The root problem seems to be that MRI `Tempfile` always wants to unlink the file, but sometimes we don't want it unlinked, and there's no public API to disable it.  A solution to that is to stop using `Tempfile`, and use `Dir::Tmpname` and a `File` instead.  Coincidentally, this happens to be how [JRuby implements Tempfile](https://github.com/jruby/jruby/blob/2c3147e674967e87cd227ed6afd379cb192008a7/lib/ruby/stdlib/tempfile.rb#L36-L58).

jjb:
@mnaberez Thanks for your attention to this. This discussion should really be in #253, because #252 only proposes changes to the test environment.

I've gone ahead and remove the commit from #253 which changed the rename behavior. Assuming that you still think it's changes are unacceptable, then I don't understand the situation. Sorry if this has already been explained by can you explain:

Is it necessary for the tempfile must be closed before it is moved?

If not, and assuming the problem is that the finalizer silently fails after the file has been successfully moved, why does it matter that the finalizer silently fails.

Thanks again for your attention to this and sorry for all the back and forth!

mnaberez:
> Is it necessary for the tempfile to be closed before it is moved?

It's operating system dependent.  It doesn't have to be closed on Unix, but it does on Windows.

    C:\Users\username>irb
    irb(main):001:0> RUBY_VERSION
    => "2.2.3"
    irb(main):002:0> RUBY_PLATFORM
    => "x64-mingw32"
    irb(main):003:0> f = File.open("foo", "w")
    => #<File:foo>
    irb(main):004:0> File.rename("foo", "bar")
    Errno::EACCES: Permission denied @ sys_fail2 - (foo, bar)
            from (irb):4:in `rename'
            from (irb):4
            from C:/Ruby22-x64/bin/irb:11:in `<main>'
    irb(main):005:0> f.close
    => nil
    irb(main):006:0> File.rename("foo", "bar")
    => 0
    irb(main):007:0>    

The original code in Ruby closes the file before moving it.  That may have been for Windows support.

chrisseaton:
> The use of ObjectSpace was introduced in RubyZip in 3b639b6 (pull request #185).

Sorry if I'm not following this discussion properly - but this is about how that commit requires `ObjectSpace` to be enabled in JRuby, isn't it? But `ObjectSpace.undefine_finalizer` doesn't require `ObjectSpace` to be enabled, does it?

headius:
Protip... Use the tmpdir lib to pick a temp location for the lib and manage
it yourself rather than hacking around tempfile? It seems like most of
tempfile's features are working against what you want.

mnaberez:
From reading the diff of 8447f0e2302989615eb0141f31911fe36130f292, it removed `ObjectSpace` but it looks like it will have the side effect of causing #185 (unexpected unlink) again.  The unlink issue could be solved by changing this code to not use `Tempfile`, as suggested [here](https://github.com/rubyzip/rubyzip/pull/252#issuecomment-158550220) and [here](https://github.com/rubyzip/rubyzip/pull/252#issuecomment-158591725).

8447f0e2302989615eb0141f31911fe36130f292 also introduces a new potential source of corruption by turning off `binmode`, see #258.

jjb:
> It's operating system dependent. It doesn't have to be closed on Unix, but it does on Windows.

ahhh, okay.

> ObjectSpace.undefine_finalizer doesn't require ObjectSpace to be enabled, does it?

whoah. I did not know that. you're right!

```
irb(main):001:0> RUBY_ENGINE
=> "jruby"
irb(main):002:0> ObjectSpace
=> ObjectSpace
irb(main):003:0> ObjectSpace.undefine_finalizer
ArgumentError: wrong number of arguments calling `undefine_finalizer` (0 for 1)
	from (irb):3:in `<eval>'
	from org/jruby/RubyKernel.java:978:in `eval'
	from org/jruby/RubyKernel.java:1291:in `loop'
	from org/jruby/RubyKernel.java:1098:in `catch'
	from org/jruby/RubyKernel.java:1098:in `catch'
	from /Users/john/.rbenv/versions/jruby-9.0.3.0/bin/irb:13:in `<top>'
```

@headius is that expected?

chrisseaton:
Yes that's expected. `JRuby.objectspace` returns false because full `ObjectSpace` is disabled, but the parts you are using as still enabled.

mnaberez:
Pull request #259 fixes the accidental unlink (#185) and this issue (#252).  It does not use `Tempfile` and does not use `ObjectSpace`.

PR #253, opened 2015-10-29
Refactor on success replace

jjb:
see also discussion in https://github.com/rubyzip/rubyzip/pull/252

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -405,14 +405,11 @@ module Zip
     def on_success_replace
       tmpfile      = get_tempfile
       tmp_filename = tmpfile.path
-      ObjectSpace.undefine_finalizer(tmpfile)
       tmpfile.close
       if yield tmp_filename
         ::File.rename(tmp_filename, name)
         ::File.chmod(@file_permissions, name) if defined?(@file_permissions)
       end
-    ensure
-      tmpfile.unlink if tmpfile
     end
 
     def get_tempfile
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -405,11 +405,11 @@ module Zip
     def on_success_replace
       tmpfile      = get_tempfile
       tmp_filename = tmpfile.path
-      tmpfile.close
       if yield tmp_filename
         ::File.rename(tmp_filename, name)
         ::File.chmod(@file_permissions, name) if defined?(@file_permissions)
       end
+      tmpfile.close
     end
 
     def get_tempfile


lib/zip/file.rb
```diff
@@ -405,14 +405,13 @@ def check_file(path)
     def on_success_replace
       tmpfile      = get_tempfile
       tmp_filename = tmpfile.path
-      ObjectSpace.undefine_finalizer(tmpfile)
-      tmpfile.close
       if yield tmp_filename
-        ::File.rename(tmp_filename, name)
+        new_file = ::File.new(name, "w")
+        new_file.write tmpfile.read
```
mnaberez: The tempfile stores all changes prior to `commit` or `close`.  Its size is unbounded so it shouldn't be loaded into memory all at once (#233, #250).

jjb: Thanks, fixed!

lib/zip/file.rb
```diff
@@ -405,14 +405,13 @@ def check_file(path)
     def on_success_replace
       tmpfile      = get_tempfile
       tmp_filename = tmpfile.path
-      ObjectSpace.undefine_finalizer(tmpfile)
-      tmpfile.close
       if yield tmp_filename
-        ::File.rename(tmp_filename, name)
+        new_file = ::File.new(name, "w")
+        IO.copy_stream tmpfile, new_file
```
mnaberez: The original code uses File.rename to move the tempfile into place, which is a wrapper around the C function rename(3) so it is fast and atomic.  The proposed change is undesirable because it is slow (reads/writes entire ZIP file again) but more importantly not atomic, so it can leave a corrupt ZIP file behind if it is interrupted.

jjb: I see. The problem I am trying to get around is:

* is `tmpfile.close` guaranteed to leave the file on the filesystem, I think it is not, hence https://github.com/rubyzip/rubyzip/pull/185
* we could solve the problem from https://github.com/rubyzip/rubyzip/pull/185 and maybe other problems by moving tmpfile.close to below the rename (as I have done in 7432e834c12e8314b84ad2d475bb55e77cb90e1b), but ``::File.rename`` on most (all?) systems uses the same file with the same inode. is `tmpfile.close` guaranteed to delete by file path and not by inode? the if so, then that's the solution. looking at the MRI implementation (https://github.com/ruby/ruby/blob/trunk/lib/tempfile.rb#L159, https://github.com/ruby/ruby/blob/trunk/lib/tempfile.rb#L203), this seems to be the case, so this could be the way to go.

So, we could remove 1ad5814fd95f46fc3504306899ccfb58c7e3479c from my PR and leave the other two.

Thoughts?

jjb:
@mnaberez ping

headius:
There's nothing here that requires JRuby's ObjectSpace features to be enabled, but our Tempfile does not use ObjectSpace for finalization.

jjb:
@headius This PR removes `ObjectSpace.undefine_finalizer(tmpfile)`, which is the only place in rubyzip where objectspace is needed in production. The rest of the uses are in tests. So with this PR, objectspace will no longer be needed for rubyzip's features.

jjb:
(I went ahead and removed the commit which changed the renaming behavior.)

mnaberez:
This PR can be closed now.  It is obsolete now that #259 has been merged.

PR #254, opened 2015-11-05
Restore 1.9.3 and add 1.9.2 to .travis.yml

agrimm:
1.9.3 was removed from .travis.yml in 0cbae14c76d698840f4c60e62bda6ce0f30e6f37 but without an explanation for doing so.

Rubyzip is described as requiring 1.9.2 or greater, without any versions being described as unsupported, so 1.9.2 and 1.9.3 should be included in the versions tested.

The main reason I'm adding them to the testing matrix is that I'm investigating a problem with rb-roo/roo and Ruby 1.9.3, and I want to check whether roo or rubyzip is the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: ruby
 sudo: false
 cache: bundler
 rvm:
+  - 1.9.2
+  - 1.9.3
   - 2.0.0
   - 2.1.5
   - 2.2.0


mnaberez:
@simonoff I work on a project that still uses 1.9.3 with RubyZip.  We would appreciate if 1.9.3 compatibility is retained.  If 1.9.3 support is to be dropped, please give notice and bump the major version.

agrimm:
Hmm ... all the CIs from 1.9.2 through to 2.2.0 failed to even build. I suspect it's not something specific to 1.9.2 or 1.9.3. Is it possible to re-run the builds?

hainesr:
I suspect support for ruby 1.9.x has been dropped because they are all out of extended maintenance support now. 1.9.2 ended on July 31, 2014, and 1.9.3 ended on February 23, 2015.

I do agree that the major version should have been bumped though, as per [semantic versioning](http://semver.org/).

simonoff:
@agrimm I don't see any reasons to do it. This versions unsupported and was not stable. Only 2.x stable branch.

PR #255, opened 2015-11-05
Allow Zip::Entry#extract to handle destinations with missing directories

weynsee:
`Zip::Entry#extract` raises an error when attempting to extract to a nested file path that is missing some directories. This PR makes sure all directories are created before extracting to the file path.

I'm not sure if this should be handled by the client (by creating the folders before calling it) but since `extract` already handles non-existent directories for directory paths I thought it would be consistent to do this for files as well.


stinnux:
I have the same problem and encountered various zip files today that are missing the directory. Highly recommend merging the PR

PR #256, opened 2015-11-10
Drop Ruby 1.9 support in changelog and gemspec

mnaberez:
Ruby 1.9 support was silently dropped in 0cbae14c76d698840f4c60e62bda6ce0f30e6f37 (see issue #254).  Support for a major Ruby version should not be dropped without notifying users.  If Ruby 1.9 support is to be dropped, please update the Rubyzip major version, changelog, and gemspec.

cc: @agrimm @hainesr @simonoff


simonoff:
I don't see any reasons in that. What it will give up? Nothing. Ruby Core team already not supported them. So IF anyone will have issues with Ruby 1.9 then we can think something. But again it's not supported software.

mnaberez:
Ruby 1.9.3 came out of official support this year but that doesn't mean every deployment of it immediately disappeared.  It was a stable Ruby version that was run by production applications for years.  Some users may still be transitioning for various reasons.

Please do one of the following: 

- Merge #254 to restore testing 1.9.3 on CI and keep supporting 1.9.3.
- Merge this pull request (#256) to communicate that you have dropped Ruby 1.9.3 support.

You have silently dropped CI testing on 1.9.3 in 0cbae14c76d698840f4c60e62bda6ce0f30e6f37.  Any change you make may break 1.9.3 and you won't know.  Users do not know you have dropped 1.9.3 support because you haven't communicated it to them in the changelog, gemspec, or version.  Users running on 1.9.3 may upgrade via `bundle update` and suddenly find their app is broken because of Rubyzip.  There's not a changelog entry to help them figure out why and the gemspec still says 1.9.3 is supported.

hainesr:
I agree that the major version number should have been bumped if support for 1.9.x was dropped. So why not do it now and keep the 1.1.x branch for fixes needed by 1.9.x if necessary?

PR #264, opened 2015-12-19
Fix TypeError: can't dup NilClass from EntrySet#dup

shhavel:
```ruby
Zip::File.open('test/data/test.xls') # => TypeError: can't dup NilClass
# from rubyzip/lib/zip/entry_set.rb:48:in `dup'
```
`EntrySet` contains 5 `nil` values:
```ruby
{ "" => nil, "" => nil, "" => nil, "" => nil, "" => nil }
```
it causes exception in ` EntrySet#dup`

Maybe `Zip::Error` should be raised instead.

```bash
$ unzip test/data/test.xls
Archive:  test/data/test.xls
warning [test/data/test.xls]:  144078 extra bytes at beginning or within zipfile
  (attempting to process anyway)
  inflating: [Content_Types].xml     
  inflating: _rels/.rels             
  inflating: theme/theme/themeManager.xml  
  inflating: theme/theme/theme1.xml  
  inflating: theme/theme/_rels/themeManager.xml.rels 
```


simonoff:
Саша дякую!

PR #277, opened 2016-02-25
Add length method to InputStream

### LINKED ISSUE #276: InputStream doesn't implement length method
I'm trying out upload files from a zip archive without splitting them apart on disk. I thought I'd be able to pass the `InputStream` directly into `UploadIO.new` and everything would be fine.

However, it looks like in order to multi-part the message body, it needs to know the `length` of the IO object. I'm happy to try adding this method, but is it even possible to know without reading the entire stream (and then rewinding it)?

zacstewart:
To make this thing act more realistically like an IO, it needs a length method.

One useful example is being able to iterate through a zip of images and upload them to an HTTP server using `UploadIO`. In order to multi-part such a POST body, the length of the IO must be known.

Fixes #276

### DIFF
--- a/lib/zip/input_stream.rb
+++ b/lib/zip/input_stream.rb
@@ -108,6 +108,11 @@ module Zip
       end
     end
 
+    def entry_size
+      return 0 if @current_entry.nil?
+      @current_entry.size
+    end
+
     protected
 
     def get_io(io_or_file, offset = 0)
--- a/lib/zip/ioextras/abstract_input_stream.rb
+++ b/lib/zip/ioextras/abstract_input_stream.rb
@@ -105,6 +105,10 @@ module Zip
       rescue EOFError
       end
 
+      def length
+        entry_size
+      end
+
       alias_method :each, :each_line
     end
   end
--- a/test/input_stream_test.rb
+++ b/test/input_stream_test.rb
@@ -179,4 +179,16 @@ class ZipInputStreamTest < MiniTest::Test
       assert_equal('$VERBOSE =', zis.read(10))
     end
   end
+
+  def test_test_entry_size
+    ::Zip::InputStream.open(TestZipFile::TEST_ZIP2.zip_name) do |zis|
+      assert_equal(zis.entry_size, 0)
+      zis.get_next_entry
+      assert_equal(zis.entry_size, 123702)
+      zis.get_next_entry
+      zis.get_next_entry
+      zis.get_next_entry
+      assert_equal(zis.entry_size, 6)
+    end
+  end
 end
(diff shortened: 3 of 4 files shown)


mnaberez:
The omission of the `length` was likely intentional because there is not a way to know the size of the decompressed data without decompressing it.  The size reported by a zip entry should be considered advisory only and can't be relied on in this way.  The entry is easily manipulated and this is how [zip bombs](https://en.wikipedia.org/wiki/Zip_bomb) work; the entry reports a small size but the data decompresses much larger. 

If you are creating the ZIP files yourself and know that the entry size is reliable, using the entry size in this way might be acceptable.  However, I don't support this as a general addition to RubyZip because it can't be used reliably on ZIP files from other sources.

zacstewart:
Are there any risks that outweigh the benefits? From my perspective, a zip with inaccurately reported entry sizes is just invalid, and with invalid data you can expect invalid behavior.

Unless there are security implications, it seems useful to me to be able to treat the stream like an IO, even if it requires some implicit trust in the archive being truthful about the decompressed size of its entries

mnaberez:
> Are there any risks that outweigh the benefits? From my perspective, a zip with inaccurately reported entry sizes is just invalid, and with invalid data you can expect invalid behavior.

 RubyZip is sometimes used in web applications, and the ZIP files being decompressed may come from user input.  [Zip bombs](https://en.wikipedia.org/wiki/Zip_bomb) are a well-known and commonly used as a denial of service attack.  The ZIP bomb works by exploiting the possible inconsistency between the size reported in the entry and the actual size the data will decompress to.  A very tiny ZIP file may expand many gigabytes and consume all the space on the disk.  Web developers can't trust ZIP files as simply "invalid" or else they will be vulnerable to these attacks.

hainesr:
Playing *Devil's Advocate* a bit here, but is it the responsibility of RubyZip to save badly designed Web applications from themselves? Or, to put it another way: is it reasonable to omit a useful, sensible and (maybe) expected feature just because people may abuse it?

I suppose a third way of looking at it might also be: There's no length attribute now, so how would a Web application make decisions on how to unpack things as they stand? At least if there was an advertised length then one could stop reading when one hit that length.

mnaberez:
> I suppose a third way of looking at it might also be: There's no length attribute now, so how would a Web application make decisions on how to unpack things as they stand?

The expected size of the decompressed data is already available from `ZipEntry#size`.  The application can call `entry.size` to get the expected size before calling `entry.get_input_stream` to get the stream.

This pull request is about adding a `length` method to the stream itself, and having it return `entry.size`.  This would allow more places that expect an IO-like object to accept the stream object.  

I think the reason `length` was not added to the stream originally is because the real size of the decompressed data can't be known without decompressing.  I don't support adding `length` for that reason and because I think it makes it easier for users to implement a bad practice.  The size reported by the entry should not be trusted blindly.

simonoff:
I don't a reasons for it. If you need such method you can wrap it in your code.

PR #298, opened 2016-08-28
Clean up empty.zip after running tests.

hainesr:
After running the tests there was a file left in the project directory,
'empty.zip', which was simply 'touch'ed to use as a null file. Switch to create
this as one of the generated test files so it's cleaned up automatically.

Also, use 'null.zip' as the name as there is already an 'empty.zip' generated
that is a zip file, but empty.


hainesr:
Sigh. The tests fail due to jruby only. I'm looking into it....

PR #300, opened 2016-09-01
Fix permissions on new zip files (#294)

### LINKED ISSUE #294: Set correct file permissions given a umask of 0027
We lock down our umask to 0027 so that files are not globally readable. When creating files via `File.open` this results in files with permissions of 0640 (i.e. `rw-r----`) but zip files created with this library end up with the very odd permission of 0637 (e.g. `rw--wxrwx`).

[I've added a failing test case to `test/file_permissions_test.rb`](https://github.com/rubyzip/rubyzip/compare/7aa3666e341b4ecf3b66321d734802a77bba962b...metavida:umask-0027) on my branch but haven't had time to actually write a solution nor to do any testing on windows.

hainesr:
This is an attempt to fix #294 in a more sensible way than was attempted, by me :-), previously.

I used the improved tests by @metavida.

To fix it I stopped the original temporary files from being created with 0600 permissions. Is this sensible? Have I missed something?

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -77,7 +77,6 @@ module Zip
           read_from_stream(f)
         end
       when create
-        @file_permissions = create_file_permissions
         @entry_set = EntrySet.new
       when ::File.zero?(file_name)
         raise Error, "File #{file_name} has zero size. Did you mean to pass the create flag?"
@@ -406,7 +405,7 @@ module Zip
       tmp_filename = create_tmpname
       if yield tmp_filename
         ::File.rename(tmp_filename, name)
-        ::File.chmod(@file_permissions, name) if defined?(@file_permissions)
+        ::File.chmod(@file_permissions, name) if @create.nil?
       end
     ensure
       ::File.unlink(tmp_filename) if ::File.exist?(tmp_filename)
@@ -415,14 +414,10 @@ module Zip
     def create_tmpname
       dirname, basename = ::File.split(name)
       ::Dir::Tmpname.create(basename, dirname) do |tmpname|
-        opts = {perm: 0600, mode: ::File::CREAT | ::File::WRONLY | ::File::EXCL}
+        opts = {mode: ::File::CREAT | ::File::WRONLY | ::File::EXCL}
         f = File.open(tmpname, opts)
         f.close
       end
-    end
-
-    def create_file_permissions
-      ::Zip::RUNNING_ON_WINDOWS ? 0644 : 0666 - ::File.umask
     end
   end
 end
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -402,23 +402,19 @@ module Zip
     end
 
     def on_success_replace
-      tmp_filename = create_tmpname
-      if yield tmp_filename
-        ::File.rename(tmp_filename, name)
-        ::File.chmod(@file_permissions, name) if @create.nil?
+      dirname, basename = ::File.split(name)
+      ::Dir::Tmpname.create(basename, dirname) do |tmp_filename|
+        begin
+          if yield tmp_filename
+            ::File.rename(tmp_filename, name)
+            ::File.chmod(@file_permissions, name) if @create.nil?
+          end
+        ensure
+          ::File.unlink(tmp_filename) if ::File.exist?(tmp_filename)
+        end
       end
-    ensure
-      ::File.unlink(tmp_filename) if ::File.exist?(tmp_filename)
     end
 
-    def create_tmpname
-      dirname, basename = ::File.split(name)
-      ::Dir::Tmpname.create(basename, dirname) do |tmpname|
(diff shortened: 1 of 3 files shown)


lib/zip/file.rb
```diff
@@ -406,7 +405,7 @@ def on_success_replace
       tmp_filename = create_tmpname
       if yield tmp_filename
         ::File.rename(tmp_filename, name)
-        ::File.chmod(@file_permissions, name) if defined?(@file_permissions)
+        ::File.chmod(@file_permissions, name) if @create.nil?
```
metavida: Why change this to `if @create.nil?` instead of `defined?(@file_permissions)`

With your removal of the `@file_permissions` setter in the `when create` branch should mean that that variable is only-ever set for the non-create path. 

I also lean towards avoiding a `.nil?` check on the `@create` variable, because the desired values for `@create` are a bit fuzzy to me, as a developer who's a bit unfamiliar with the code. The code comment for the `initialize` method suggests that `@create` should be either `nil` or `true`, but the constant is `CREATE = 1` 😃

lib/zip/file.rb
```diff
@@ -415,15 +414,11 @@ def on_success_replace
     def create_tmpname
       dirname, basename = ::File.split(name)
       ::Dir::Tmpname.create(basename, dirname) do |tmpname|
-        opts = {perm: 0600, mode: ::File::CREAT | ::File::WRONLY | ::File::EXCL}
+        opts = {mode: ::File::CREAT | ::File::WRONLY | ::File::EXCL}
```
metavida: At first I was worried that removing the `perm: 0600` bit of code was going to result in a security hole (i.e. the tempfile being created in a shared directory like `/tmp/` with permissions where another user could potentially read the file), *but* after much testing I validated that there is **not** a security risk! Woot! It became clear to me that because we're passing the basename to `Dir::Tmpname`, the tempfile is *not* being created in a shared space, unless the person using this gem specifically points `File::Zip.open` at a shared directory.

*Note:* this comment is just for posterity, in case a future dev has concerns about this same potential security flaw.

metavida:
@hainesr, looks sensible to me! There's [one note suggesting not using a `@create.nil?` check](https://github.com/rubyzip/rubyzip/pull/300#discussion_r77461336) but that's up to you. You've got my 👍  to merge this! Thanks for the fix and for cleaning up my unit tests too!

lib/zip/file.rb
hainesr: The main reasons I stopped using the `@file_permissions` variable here are:
* The logic for whether we need to change the permissions here is based upon whether the file is created or not, not whether there exists some file permissions. So,
* If someone decided to use the `@file_permissions` variable for something else, we might get an odd bug here.

I agree with your concerns about the `@create` variable though. I'll look at whether I can make its use consistent.

lib/zip/file.rb
hainesr: Yes, my original plan to fix this was to make sure the file is created in the same place as its eventual destination - then I noticed that is exactly what is already being done!

hainesr:
Thanks for your notes @metavida! I'll see if I can fix the usage of the `@create` variable in a sensible way to make that bit a little more consistent. I definitely agree with your concerns with it.

metavida:
Maybe something as simple as switching to a private method like this would do the trick:

```
def create?
  # In order to not break backwards comparability
  # allow any truthy value to work. 
  @create
end
``` 

Then updating the code comment for `initislize` to reference the `CREATE` constant, just so that it matches the rest of the code samples.

hainesr:
I've been playing with this today. I added a couple of debug `puts` to see what values `@create` is ever set to and got some *very* odd results!

There were many cases where `@create` was set to `{:mode=>193}`! It took me ages to find out why this was but I eventually tracked it down to here: https://github.com/rubyzip/rubyzip/blob/master/lib/zip/file.rb#L419

```ruby
f = File.open(tmpname, opts)
```

This, of course, calls the `Zip::File.open` method, passing `{:mode=>193}` into the `@create` variable, but I *think* it's supposed to be calling `::File.open`. Changing that line to

```ruby
f = ::File.open(tmpname, opts)
```
still passes all the tests and I think will make it easier to fix `@create` to be more consistent.

I'll add a commit to this pull request soon.

hainesr:
Hmm, actually, as that file is simply opened and then immediately closed, maybe the only reason that code is there was to set the permissions and raise an error if the file already existed?

Given we don't want the permissions to be set any more, and the file will not exist as we're generating a new random filename, then maybe it could be refactored out completely.

I'll have a go and see what happens.

metavida:
Awesome! Best case scenario for a bug fix: less application code to maintain and more test coverage! Thanks again for the time you're putting in to maintain this gem. It's been quite handy for us!

hainesr:
Thanks @metavida.

Unfortunately I can't merge this myself as I'm not a member of the rubyzip project, but hopefully someone will be along soon who can!

hainesr:
Ping?

PR #301, opened 2016-09-01
Update travis config

hainesr:
This should be fairly uncontroversial.

I've dropped support for jruby 1.9 mode seeing as we no longer test ruby 1.9 any more.


hainesr:
Would be good to see this merged as broken builds ruin confidence in any other updates.

hainesr:
Ping?

PR #325, opened 2017-04-14
Save temporary files to a temporary directory

aeroastro:
I would like to save temporary files to a temporary directory.

Because the primary objective of `@temp_file` in `Zip::StreamableStream` is to temporarily hold data of an entry as a file, the upper directory does not need to be the current directory.

This patch set the directory to default `Dir.tmpdir`. As a result, we can stop temporary files from filling up current directory.

c.f. http://ruby-doc.org/stdlib-2.4.1/libdoc/tempfile/rdoc/Tempfile.html

### DIFF
--- a/lib/zip/streamable_stream.rb
+++ b/lib/zip/streamable_stream.rb
@@ -5,7 +5,7 @@ module Zip
       dirname = if zipfile.is_a?(::String)
                   ::File.dirname(zipfile)
                 else
-                  '.'
+                  nil
                 end
       @temp_file = Tempfile.new(::File.basename(name), dirname)
       @temp_file.binmode


aeroastro:
Should I maintain or increase coverage for this patch to be merged?

Since this patch implements correct (or recommended) arguments to Ruby built-in library, IMHO there seems to be no strong requirements to keep the coverage.

Also, the coverage decrease from https://coveralls.io/builds/10052092 (recent master push) to https://coveralls.io/builds/12143274 (this branch) is due to increased number of relevant lines from 2031 to 2032. This is actually triggered by counting `end` line as relevant lines here https://coveralls.io/builds/12143274/source?filename=lib%2Fzip%2Fstreamable_stream.rb#L9. It seems like there are some issues on coverage calculation side.


I just would like to stop temporary files from filling up application's root directory. 🐱

aeroastro:
@simonoff 

Hello.

It's been a while since I submitted this PR, and I'm wondering if there is anything I could do on my end. I totally understand you are busy and require more time to consider, it is much appreciated if you could give me a response.

aeroastro:
Thank you very much!!!

PR #326, opened 2017-04-17
Make naming on README more consistent

aeroastro:
I would like to make naming more consistent and understandable especially on README so that new developers can easily get familiar with rubyzip.

IMHO, I would like to rename variables on README in the following way.

* io => zipfile
    * First example uses `zipfile` and it is an object of Zip::File
    * Since Zip::File uses IO-like objects in some cases, this change could reduce some confusion.
* os => f
    * Second example uses `f` and it is an object of Tempfile

Any better ideas are welcome.


aeroastro:
I have resolved conflict.

aeroastro:
Thank you very much!!!

PR #335, opened 2017-08-02
add bundle install to README.md instructions

seth-at-at:


README.md
```diff
@@ -32,6 +32,8 @@ Or in your Gemfile:
 
 ```ruby
 gem 'rubyzip'
+
+bundle install
```
aeroastro: You can see the `Or in your Gemfile:` above this code block.
Since the content in this code block should be equivalent of that in Gemfile, IMHO, I think it would be preferred to mention `bundle install` outside the code block in the following way as bundler template does.

```
And then execute:

    $ bundle install

```

https://github.com/bundler/bundler/blob/6a3b09b2292405cab6b5f4d63004b5c8b782b469/lib/bundler/templates/newgem/README.md.tt

Because I am not the owner of this repository, I would like to hear from @simonoff.

simonoff: I think in any case you need to do a bundle install. So no reason to say it again for user.

PR #355, opened 2018-02-23
Fix regression caused by automatic Rubocop fixes

pnomolos:
`Zip::Entry.empty?` doesn't exist, but Rubocop suggested it as a fix instead of `size == 0`.  This particular instance only fails under very certain circumstances.  I have a file that causes it to fail but it's an proprietary internal file that's automatically generated by one of our vendors so I'm unable to share it.

I only fixed this one specific instance, there may be other regressions caused by the Rubocop cleanup so perhaps the entire commit should be double-checked.


danielrhodes:
Came here to patch the same thing! @olleolleolle @simonoff Could one of you merge this?

Issue #236, opened 2015-06-15
zip without compression extracts to empty files


ericcj:
i created a zip using https://github.com/pixelglow/zipzap with compress:NO on the entries and it unzips fine using the normal unzip command line utility (Info-ZIP) and Archive Utility on my os x laptop.  however, when i try to extract it with rubyzip it creates zero-byte files for all the entries.  here's what one of the entries and its input stream looks like.  sysread on that stream always returns empty string. 

```
#<Zip::Entry:0x007f9abcab60e0 @local_header_offset=99599183, @local_header_size=nil, @internal_file_attributes=0, @external_file_attributes=2175008768, @header_signature=33639248, @version_needed_to_extract=10, @version=30, @ftype=:file, @filepath=nil, @gp_flags=2056, @follow_symlinks=false, @restore_times=true, @restore_permissions=false, @restore_ownership=false, @unix_uid=nil, @unix_gid=nil, @unix_perms=420, @dirty=false, @fstype=3, @zipfile="/Users/ej/17ba3bfa-139f-11e5-9c00-2a828cf49bc3.zip", @name="C45664C1-ACC5-413E-9534-C7BF05DC5338.jpg", @comment="", @extra={}, @compressed_size=2239128, @crc=375996240, @compression_method=0, @size=2239128, @time=2015-06-15 16:42:50 -0400, @last_mod_time=34137, @last_mod_date=18127, @name_length=40, @extra_length=0, @comment_length=0>

[...]

PR #358, opened 2018-04-03
Handle stored files with general purpose bit 3 set

idoru:
In these cases, the `@current_entry` is unpacked from the local headers, but the size/compressed size/crc are all set to 0, which causes the `PassThruDecompressor` to read no bytes instead of the entry size.

This change makes `Entry#get_input_stream` pass the `Entry` itself down to `Zip::InputStream` so that when creating the Decompressor, the correct number of bytes can be passed down if general purpose bit 3 is set.

May be related to #236 

Signed-off-by: Sam Coward <scoward@pivotal.io>


sclevine:
@simonoff @olleolleolle any feedback? We'd like to use rubyzip in Cloud Foundry's Cloud Controller, but we're blocked on this PR.

Issue #192, opened 2014-11-20
IO class is not properly supported for streaming.


weshatheleopard:
Change the code at the following location ( https://github.com/rubyzip/rubyzip/blob/05916bf89181e1955118fd3ea059f18acac28cc8/test/file_test.rb#L538 ) to be as follows:

```
File.open(zname, 'rb') { |f|
  Zip::File.open_buffer(f) do |zipfile|
  end
}
```

According to the documentation, IO object is valid in this situation. However, the modified test fails:

```
ZipFileTest#test_streaming:
ArgumentError: wrong number of arguments (0 for 1..2)
    /mnt/var/git/rubyzip/lib/zip/output_stream.rb:32:in `reopen'
    /mnt/var/git/rubyzip/lib/zip/output_stream.rb:32:in `initialize'
    /mnt/var/git/rubyzip/lib/zip/output_stream.rb:59:in `new'
    /mnt/var/git/rubyzip/lib/zip/output_stream.rb:59:in `write_buffer'
    /mnt/var/git/rubyzip/lib/zip/file.rb:317:in `write_buffer'
    /mnt/var/git/rubyzip/lib/zip/file.rb:128:in `open_buffer'
    /mnt/var/git/rubyzip/test/file_test.rb:542:in `block in test_streaming'
    /mnt/var/git/rubyzip/test/file_test.rb:541:in `open'
    /mnt/var/git/rubyzip/test/file_test.rb:541:in `test_streaming'
```

Also see related #177.

jdleesmiller:
I think this is the same as https://github.com/rubyzip/rubyzip/issues/146 and now fixed.

Issue #146, opened 2014-03-21
Error when calling Zip::File.open_buffer(io)


davidblondeau:
Just updated from 1.1.0 to 1.1.2 and it seems like the call to reopen in OutputStream is causing an issue.

If you run the following code after replacing "something.zip" with the path to an actual file, you get the error that follows.

```
File.open("something.zip") do |io|
  io.set_encoding(Encoding::BINARY)
  Zip::File.open_buffer(io) do |zip_io|
    // left empty on purpose
  end
end
```

```
ArgumentError: wrong number of arguments (0 for 1..2)
    from vendor/ruby/1.9.1/gems/rubyzip-1.1.2/lib/zip/output_stream.rb:32:in `reopen'
    from vendor/ruby/1.9.1/gems/rubyzip-1.1.2/lib/zip/output_stream.rb:32:in `initialize'
    from vendor/ruby/1.9.1/gems/rubyzip-1.1.2/lib/zip/output_stream.rb:59:in `new'
    from vendor/ruby/1.9.1/gems/rubyzip-1.1.2/lib/zip/output_stream.rb:59:in `write_buffer'
    from vendor/ruby/1.9.1/gems/rubyzip-1.1.2/lib/zip/file.rb:316:in `write_buffer'
    from vendor/ruby/1.9.1/gems/rubyzip-1.1.2/lib/zip/file.rb:128:in `open_buffer'
    from (irb):21:in `block in irb_binding'
    from (irb):19:in `open'
    from (irb):19
    from /Users/work/.rvm/rubies/ruby-1.9.3-p484/bin/irb:12:in `<main>'
```

jdleesmiller:
I think this was fixed some time ago, but to be safe I have added a test case, which presently passes:

https://github.com/rubyzip/rubyzip/blob/eeef5073d58253e2044dbf81d1b205efd590b59a/test/file_test.rb#L134-L141

So, I'll close this out.

PR #360, opened 2018-04-04
Fix #280 - `open_buffer` mangles the content of the buffer it is given.

### LINKED ISSUE #280: `open_buffer` mangles the content of the buffer it is given.
Here, you can use absolutely any ZIP file for `test.zip`:

```
require 'zip'
buf = nil
File.open('test.zip', "rb") { |f| buf = f.read }
File.open("before.zip", "wb") { |f| f << buf } 
Zip::File.open_buffer(buf) { |zf| nil # Do nothing at all!!! }
File.open("after.zip", "wb") { |f| f << buf }
```

Note that `after.zip` differs from `before.zip`, despite the fact that there's absolutely no reason to be changed.

However, if you do:

```
File.open('test.zip', "rb") { |f| buf = f.read }
Zip::File.open_buffer(buf.freeze) { |zf| zf.entries.first }
File.open("after.zip", "wb") { |f| f << buf }
```

then everything works perfectly, `buf` is not damaged and nothing crashes.

hainesr:
Things are now more carefully set up, and if a buffer is passed in which represents a file that already exists then this is taken into account. All initialization is now done in File.new, rather than being split between there and File.open_buffer.

This has also needed a bit of a re-write of Zip::File.initialize. I've tried to bring some logic to it as a result, and have added comments to explain what is now happening.

This PR includes some other related fixes too. All tests pass.

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -64,9 +64,9 @@ module Zip
 
     # Opens a zip archive. Pass true as the second parameter to create
     # a new archive if it doesn't exist already.
-    def initialize(file_name, create = false, buffer = false, options = {})
+    def initialize(path_or_io, create = false, buffer = false, options = {})
       super()
-      @name    = file_name
+      @name    = path_or_io.respond_to?(:path) ? path_or_io.path : path_or_io
       @comment = ''
       @create  = create ? true : false # allow any truthy value to mean true
       if !buffer && ::File.size?(@name)
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -134,12 +134,10 @@ module Zip
           raise "Zip::File.open_buffer expects a String or IO-like argument (responds to #{IO_METHODS.join(', ')}). Found: #{io.class}"
         end
 
-        if io.is_a?(::String)
-          io = ::StringIO.new(io)
-        elsif io.respond_to?(:binmode)
-          # https://github.com/rubyzip/rubyzip/issues/119
-          io.binmode
-        end
+        io = ::StringIO.new(io) if io.is_a?(::String)
+
+        # https://github.com/rubyzip/rubyzip/issues/119
+        io.binmode if io.respond_to?(:binmode)
 
         zf = ::Zip::File.new(io, true, true, options)
         return zf unless block_given?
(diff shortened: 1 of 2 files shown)


test/file_test.rb
```diff
@@ -107,13 +114,14 @@ def test_open_buffer_with_stringio
   def test_close_buffer_with_stringio
     string_io = StringIO.new File.read('test/data/rubycode.zip')
     zf = ::Zip::File.open_buffer string_io
-    assert(zf.close || true) # Poor man's refute_raises
+    assert_nil zf.close
   end
 
   def test_close_buffer_with_io
     f = File.open('test/data/rubycode.zip')
```
jdleesmiller: Note: this is similar (but not identical) to the test cases in https://github.com/rubyzip/rubyzip/issues/146 and https://github.com/rubyzip/rubyzip/issues/192, so they can probably be closed once we sort this out.

lib/zip/file.rb
```diff
@@ -64,24 +64,38 @@ class File < CentralDirectory
 
     # Opens a zip archive. Pass true as the second parameter to create
     # a new archive if it doesn't exist already.
-    def initialize(file_name, create = false, buffer = false, options = {})
+    def initialize(path_or_io, create = false, buffer = false, options = {})
       super()
-      @name    = file_name
+      @name    = path_or_io.respond_to?(:path) ? path_or_io.path : path_or_io
```
jdleesmiller: I think this is good... but it seems a bit optional. Was it aimed at solving a particular problem?

hainesr:
Thanks for picking this up! It's been a while since I looked at this, but I'll try and find time to get it all back in my head and respond properly soon. As I remember it though, it turned into one of those problems that uncovered all sorts of other problems, and was originally a piece of code that was rather unclear as to what was going on 😬

jdleesmiller:
I reviewed this again today, and I think it is good to merge. Thanks again for the fix.

hainesr:
Sorry for not getting round to looking at it again myself; a change of role at work has left little time for extra-curricular coding recently, but I hope to get back into it again soon.

Thanks for merging, and thanks for picking up maintaining this gem @jdleesmiller.

jdleesmiller:
Definitely get that :) I was able to progress this mainly because we had a hackathon day at my company, Overleaf. I will work it up into a new release.

hainesr:
Oh cool. I'm a big fan of Overleaf!

Issue #364, opened 2018-05-17
Document how to create a reproducible ZIP file


guss77:
I think it will be useful to document how to create a reproducible ZIP file - i.e. a ZIP file that has identical content as long as the source files have not changed. 

This is very useful for build services and all kinds of cache management services, and its pretty easy to do - you just have to zero out the time field.

[See here for a discussion](https://content.pivotal.io/blog/barriers-to-deterministic-reproducible-zip-files)

GrimmDaddy:
zip -X -r foo1.zip foo; sleep 5; touch foo/*; zip -X -r foo2.zip foo; cmp foo1.zip foo2.zip

DanRathbun:
If PR #365 was rejected and closed, should this issue also be closed for v3.x ?
(It might have been added to the v2.x docs, but would then likely need to have been removed for 3.0.)

guss77:
While at that time, the close comment at #365 made sense to me, I think there is still a use case for forcing the file time to zero - to handle cases where the last modified timestamp changes without the file content changing: some processes can cause multiple source copies to have differing time stamps for the same content.

Its not high priority for me as I'm already doing what I need to be doing, but I thought that other people may be interested in documenting the process.

PR #365, opened 2018-05-17
Added reproducible ZIP example

guss77:
Fixes issue #364

Also see here for a discussion: https://content.pivotal.io/blog/barriers-to-deterministic-reproducible-zip-files

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,22 @@ end.string
 
 This is an experimental feature and the interface for encryption may change in future versions.
 
+### Create reproducible zip files
+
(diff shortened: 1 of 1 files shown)


tilsammans:
@simonoff any chance this might get merged?

GrimmDaddy:
>  zip -X -r foo1.zip foo; sleep 5; touch foo/*; zip -X -r foo2.zip foo; cmp foo1.zip foo2.zip

hainesr:
Thanks for the PR.
The default in Rubyzip is now (finally) to record the last modified time of a file as it's added to the archive, and then preserve that time when extracting. So I think this note is no longer needed. Please raise an issue if that's not the case - I am interested in making sure Rubyzip can be used for reproducibility purposes.

Issue #286, opened 2016-03-25
Can't produce zip with STORE method (uncompressed data)


darekrusin:
When I pack a zip using WinRar for example and select the "store" method for an uncompressed archive, the algorithm used is reported as "store" and the file size is only a few hundred bytes larger than the compressed data.

But using rubyzip, even when I select Zlib::NO_COMPRESSION something weird happens:
- the file indeed has no compression (I can see parts of the original data readable inside), but the resulting zip size is larger, it gains about 8 KB on a 60 MB file
- the file being zipped is not stored as a raw blog inside the zip but it's chunked with some kind of checksums mixed in
- the algorithm reported is "deflate" instead of "store"

This is a problem for situation where a truly uncompressed zip is needed, with a raw binary version of the original file inside.

Am I doing something wrong? Is there a way to force the gem to work this way, or is this a bug?

julik:
`Zlib::NO_COMPRESSION` is a setting in Zlib that is probably not usable with ZIPs. To use the stored mode you have to use the STORED mode in RubyZip. We use it in our streaming Zip archiving library but I do not remember whether RubyZip exposes it natively. I think it does.
TL;DR - `Zlib::NO_COMPRESSION` is actually _also_ a sort of deflate, so what you are seeing makes sense.

hainesr:
I've forced a STORE like this in the past: https://github.com/myGrid/ruby-zip-container/blob/master/lib/zip-container/file.rb#L79

I'm pretty sure that does what you want.

atz:
Same problem, as affecting our project here:
https://github.com/sul-dlss/preservation_catalog/issues/737#issuecomment-388504059

See the expected output (like `zip -vr0X -s 10g ...`):
```
$ unzip -lv example.zip
Archive:  example.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
       0  Stored        0   0% 04-20-2018 12:06 00000000  jj925bx9565/v0001/
       0  Stored        0   0% 04-20-2018 12:06 00000000  jj925bx9565/v0001/data/
       0  Stored        0   0% 04-20-2018 12:06 00000000  jj925bx9565/v0001/data/content/
 6508496  Stored  6508496   0% 04-20-2018 12:06 ac341bd4  jj925bx9565/v0001/data/content/PC0170_s3_4th_of_July_2010-07-04_095432_0003.jpg
[...]

PR #366, opened 2018-05-20
Zip::File.add_stored() to add uncompressed files.

hainesr:
Adding uncompressed files to a zip archive can be overly complex, so this convenience method makes it easier. Examples of the confusion caused by the current situation are in #286 and https://github.com/sul-dlss/preservation_catalog/issues/737.


hainesr:
I've rebased this onto head and all checks now pass. Do you think it's worth a look @jdleesmiller?

Issue #369, opened 2018-06-14
Several directory traversal vulnerabilities


tuzovakaoff:
## Overview
This issue is similar to already closed https://github.com/rubyzip/rubyzip/issues/315 but I found two ways to bypass that fix.

You can find files for tests in  https://github.com/tuzovakaoff/zip_path_traversal

##### rubyzip.rb

```ruby
require 'zip'
first_arg, *the_rest = ARGV

Zip::File.open(first_arg) do |zip_file|
  zip_file.each do |entry|
    puts "Extracting #{entry.name}"
    entry.extract(entry.name)
  end
end
```

#### UnZip strips absolute path
```bash
> ls
absolutepath.zip        rubyzip.rb              symlink.zip

> unzip absolutepath.zip
Archive:  absolutepath.zip
warning:  stripped absolute path spec from /tmp/
   creating: tmp/
warning:  stripped absolute path spec from /tmp/file.txt
  inflating: tmp/file.txt

> ls
absolutepath.zip	symlink.zip
rubyzip.rb		tmp
```

#### rubyzip extracts files with absolute path 
``` bash
> ls
absolutepath.zip        rubyzip.rb              symlink.zip

> ruby rubyzip.rb absolutepath.zip
Extracting /tmp/
Extracting /tmp/file.txt

> ls
absolutepath.zip        rubyzip.rb              symlink.zip

> ls /tmp
file.txt

```

#### UnZip extracts  only symbolic link
```
> unzip symlink.zip
Archive:  symlink.zip
    linking: path                    -> ../../../../../tmp
checkdir error:  path exists but is not directory
                 unable to process path/file.txt.
finishing deferred symbolic links:
  path                   -> ../../../../../tmp


> ls -l path
lrwxr-xr-x  1 user  group  18 Jun 14 21:19 path -> ../../../../../tmp
```

#### rubyzip extracts symbolic link and puts  file into `/tmp` folder
```
> ruby rubyzip.rb symlink.zip 
Extracting path
Extracting path/file.txt

> ls -l path
lrwxr-xr-x  1 user  group  18 Jun 14 21:22 path -> ../../../../../tmp

>ls /tmp
file.txt

```

## Vulnerable version and test environment
```bash
> uname -rsv
Darwin 17.6.0 Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64

> ruby -v
[...]

PR #371, opened 2018-07-01
Fix CVE-2018-1000544 

bdewater:
Attempt to fix https://github.com/rubyzip/rubyzip/issues/369

Not sure if there's a better way to handle Errno::ENOENT in the symlink case, open to suggstions.


lib/zip/entry.rb
```diff
@@ -147,14 +147,21 @@ def next_header_offset #:nodoc:all
...
-
-      if @name.squeeze('/') =~ /\.{2}(?:\/|\z)/
+    def extract(dest_path = nil, &block)
+      if dest_path.nil? && Pathname.new(@name).absolute?
+        puts "WARNING: skipped absolute path in #{@name}"
+        return self
+      elsif @name.squeeze('/') =~ /\.{2}(?:\/|\z)/
         puts "WARNING: skipped \"../\" path component(s) in #{@name}"
         return self
+      elsif symlink? && get_input_stream.read =~ %r{../..}
```
JackMc: Could you add a test that this still works for non-malicious symlinks? It doesn't appear there is one.

The comment on `get_input_stream`'s definition seems to also say that it doesn't work too great for symlinks. We should make sure we're not adding any undefined behaviour: https://github.com/rubyzip/rubyzip/blob/master/lib/zip/entry.rb#L484

tuzovakaoff: Symlink could also links to absolute path
You can find examples of such zip files in https://github.com/jwilk/path-traversal-samples

zeha:
Given this doesn't look too broken, can we please have that merged and a new release stamped with it? Thanks :-)

lib/zip/entry.rb
```diff
@@ -147,14 +147,21 @@ def next_header_offset #:nodoc:all
...
     # Extracts entry to file dest_path (defaults to @name).
-    def extract(dest_path = @name, &block)
-      block ||= proc { ::Zip.on_exists_proc }
-
-      if @name.squeeze('/') =~ /\.{2}(?:\/|\z)/
+    def extract(dest_path = nil, &block)
+      if dest_path.nil? && Pathname.new(@name).absolute?
+        puts "WARNING: skipped absolute path in #{@name}"
+        return self
+      elsif @name.squeeze('/') =~ /\.{2}(?:\/|\z)/
```
zeha: Should this not also gain the `dest_path.nil?` check?

zeha:
Example in README calls .extract() with the name from the zipfile, leading to the same vulnerability in code that gets copy-pasted from the README.

drn:
Our team would love to see this merged and released as well. This has started to show up in our automated vulnerability analysis (https://hakiri.io/technologies/rubyzip/issues/1e6c19ff65e713) and our builds are red as a result.

Happy to help get this over the line if any additional work is needed. Please let me know!

bdewater:
I can address feedback to this PR - @simonoff have you had a chance to look at this?

Right now do not have the bandwidth available to look at/add tests for all the examples in https://github.com/jwilk/path-traversal-samples - so if somebody feels like taking a crack at it you're most welcome 🙂

simonoff:
Tests is failed. Need to have a green build to make it merged.

bdewater:
@simonoff like I [mentioned](https://github.com/rubyzip/rubyzip/pull/371#issuecomment-401630960) the test failures seem unrelated - can you try to rebuild?

reedloden:
This was just added to [ruby-advisory-db](https://github.com/rubysec/ruby-advisory-db/pull/346), so would love for this PR to be finalized, landed, and released.

vipulnsward:
I looked at why the jruby builds were failing. Its after an upgrade to 9.2 affecting the crc32 checksum in Encryption tests. 
We can ignore failures on JRuby, since its a difference in MRI and JRuby implementation for now.

bdewater:
Thanks @vipulnsward - I've pulled your commit in to get this PR green.

jdleesmiller:
Based on https://github.com/rubyzip/rubyzip/pull/371#discussion_r199429562 and https://github.com/rubyzip/rubyzip/pull/371#issuecomment-414782527 I added tests for https://github.com/jwilk/path-traversal-samples

Tests are in: https://github.com/jdleesmiller/rubyzip/commit/9c468f30f38d09451e5a65edfff277cfe381fd49

Good news: most pass.

Bad news: two don't pass: `test_directory_symlink` and `test_two_directory_symlinks_b`:

```
  1) Failure:
PathTraversalTest#test_two_directory_symlinks_b [/Users/john/ex/rubyzip/test/path_traversal_test.rb:85]:
Expected false to be truthy.

  2) Failure:
PathTraversalTest#test_directory_symlink [/Users/john/ex/rubyzip/test/path_traversal_test.rb:65]:
Expected false to be truthy.
```

I have not yet attempted to fix the bugs, and I'm not sure I know enough about this library to do that, but I thought I should at least report the test results.

PR #374, opened 2018-08-13
Add test build with frozen strings

miry:
In ruby 3 option `--enable-frozen-string-literal` will be enabled by
default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ rvm:
   - 2.2.9
   - 2.3.6
   - 2.4.3
-  - 2.5.0
+  - 2.5.1
   - ruby-head
 matrix:
   include:
@@ -17,6 +17,9 @@ matrix:
     jdk: openjdk7
   - rvm: jruby-head
     jdk: oraclejdk8
+  - rvm: ruby-head
+    env:
+      - RUBYOPT="--enable-frozen-string-literal"
   - rvm: rbx-3
     env:
       - RUBYOPT="-rbundler/deprecate"
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,9 @@ matrix:
   - rvm: ruby-head
     env:
       - RUBYOPT="--enable-frozen-string-literal"
+  - rvm: ruby-head
+    env:
+      - RUBYOPT=""
   - rvm: rbx-3
     env:
       - RUBYOPT="-rbundler/deprecate"
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,6 @@ matrix:
   - rvm: ruby-head
     env:
       - RUBYOPT="--enable-frozen-string-literal"
-  - rvm: ruby-head
-    env:
-      - RUBYOPT=""
   - rvm: rbx-3
     env:
       - RUBYOPT="-rbundler/deprecate"
--- a/lib/zip/ioextras/abstract_input_stream.rb
+++ b/lib/zip/ioextras/abstract_input_stream.rb
@@ -11,13 +11,13 @@ module Zip
         super
         @lineno        = 0
         @pos           = 0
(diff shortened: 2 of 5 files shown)


lib/zip/ioextras/abstract_input_stream.rb
```diff
@@ -11,26 +11,27 @@ def initialize
         super
         @lineno        = 0
         @pos           = 0
-        @output_buffer = ''
+        @output_buffer = String.new
```
bdewater: Same encoding issue.

lib/zip/ioextras/abstract_input_stream.rb
```diff
@@ -11,26 +11,27 @@ def initialize
...
       end
 
       attr_accessor :lineno
       attr_reader :pos
 
       def read(number_of_bytes = nil, buf = '')
+        buffer = buf.dup
```
bdewater: Why don't you make `buf` a mutable `String.new('')` instead?

lib/zip/ioextras/abstract_input_stream.rb
```diff
@@ -76,7 +77,7 @@ def gets(a_sep_string = $/, number_of_bytes = nil)
         while (match_index = @output_buffer.index(a_sep_string, buffer_index)).nil? && !over_limit
           buffer_index = [buffer_index, @output_buffer.bytesize - a_sep_string.bytesize].max
           return @output_buffer.empty? ? nil : flush if input_finished?
-          @output_buffer << produce_input
+          @output_buffer += produce_input
```
bdewater: Same appending issue as mentioned above.

lib/zip/ioextras/abstract_input_stream.rb
```diff
@@ -90,7 +91,7 @@ def ungetc(byte)
 
       def flush
         ret_val        = @output_buffer
-        @output_buffer = ''
+        @output_buffer = String.new
```
bdewater: Same encoding issue.

.travis.yml
```diff
@@ -17,6 +17,9 @@ matrix:
     jdk: openjdk7
   - rvm: jruby-head
     jdk: oraclejdk8
+  - rvm: ruby-head
+    env:
+      - RUBYOPT="--enable-frozen-string-literal"
```
bdewater: Why not put the `# frozen_string_literal: true` magic comment a top of every file so that every Ruby 2.3+ user can reap the benefits? I doubt a lot of people can their Ruby app with this RUBYOPT flag since not every gem is ready for this yet.

test/filesystem/file_nonmutating_test.rb
```diff
@@ -55,7 +55,7 @@ def test_open_read
                    f.readline.chomp)
     end
     assert(blockCalled)
-    @zip_file.dir.chdir '/'
+    @zip_file.dir.chdir String.new('/')
```
bdewater: This change is not needed.

.travis.yml
miry: @bdewater I don't think to update all files in gem, without automatic rubocop/lint checking is useful. It would inconsistent file to file.

I wanted to check that gem is ready,  even without force people who uses gem to fix code. I came to this PR exactly, because enabled in all file of a project `# frozen_string_literal: true`. I don't think I fixed all problems. But at least it would be in the matrix.

miry: This record is optional. And it is included ruby head tests without any env.

lib/zip/ioextras/abstract_input_stream.rb
miry: reverted.

miry:
@bdewater thanks for your comments. I updated PR.

miry:
closing because there were no action for last 30 days.

PR #375, opened 2018-08-26
Move jruby to allow failures matrix

vipulnsward:
See: https://github.com/jruby/jruby/issues/5290

Ref: https://github.com/jruby/jruby/pull/5088


jdleesmiller:
Thanks! I am helping Alex out with maintenance now. It looks like this change already landed in #371. I will merge this to make it official. I think we can probably move jruby back now, because the problem has since been fixed.

PR #387, opened 2019-01-23
Add Changelog for 1.2.2 (#378)

### LINKED ISSUE #378: Changelog is missing the details about the 1.2.2 release
Title says it all, there are no details on the Changelog about what changes were included in the 1.2.2 release.

Thanks!

aspiers:
1.2.2 was already released in #376, so unfortunately this is too late for inclusion in that, but at least future releases will have it.

This is just a list of the titles of all non-merge commits since 1.2.1, so it won't be as concise or readable a summary as for previous releases, but it's better than nothing, and anyone is welcome to volunteer to condense it further.

Closes #378.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,40 @@
+1.2.2
+=====
+
+* Expand from root rather than current working directory
+* Disable symlinks and check for path traversal
+* Consolidate path traversal tests
+* Add jwilk's path traversal tests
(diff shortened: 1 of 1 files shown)


jdleesmiller:
... eventually. It doesn't look like I have write access yet, even though I'm now an org member. Watch this space.

PR #390, opened 2019-02-28
fixed FrozenError caused by frozen-string-literal

taichi-ishitani:
Hi,

I got FrozenError when I used rubyzip gem with `--enable-frozen-string-literal` option.
```
FrozenError:
       can't modify frozen String, created at C:/Ruby/Ruby25-x64/lib/ruby/gems/2.5.0/gems/rubyzip-1.2.2/lib/zip/entry.rb:605
```
This PR is to fix the above error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,10 @@ matrix:
     - rvm: jruby-head
     - rvm: jruby
 before_install:
-  - gem update --system
-  - gem install bundler
+  - "if $(ruby -e 'exit(RUBY_VERSION >= \"2.3.0\")'); then gem update --system; fi"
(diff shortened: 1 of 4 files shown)


jdleesmiller:
Linking this up: https://github.com/rubyzip/rubyzip/pull/374 also tried to address this. Neither PR is a superset of the other, so I'm not sure either one addressed the issue fully, but it is probably all still helping.

I think the best way forward is to get rubocop up to date and enforce the `# frozen_string_literal: true` magic comments in all the rubyzip files, so it's not dependent on end user flags.

taichi-ishitani:
https://github.com/Invoca/magic_frozen_string_literal

I think this gem can help you to add the magic comment.

Issue #384, opened 2018-10-02
Tests for the Zip Slip vuln raise alarms by ClamAV antivirus...


MaximeDerche:
Hello,

Since the 21st Sept. I get the following messages from my daily ClamAV scan:

/usr/share/logstash/vendor/bundle/jruby/2.3.0/gems/rubyzip-1.2.2/test/data/path_traversal/tuzovakaoff/symlink.zip: Sanesecurity.Malware.27384.ZipHeur.ZipSlip.UNOFFICIAL FOUND
/usr/share/logstash/vendor/bundle/jruby/2.3.0/gems/rubyzip-1.2.2/test/data/path_traversal/relative1.zip: Sanesecurity.Malware.27384.ZipHeur.ZipSlip.UNOFFICIAL FOUND
/usr/share/logstash/vendor/bundle/jruby/2.3.0/gems/rubyzip-1.2.2/test/data/path_traversal/jwilk/relative2.zip: Sanesecurity.Malware.27384.ZipHeur.ZipSlip.UNOFFICIAL FOUND

This is because the Sanesecurity signature base gets those three files as vulnerable to a known risk.

I understand those files come from the commit d07b13a6cf0a413e010c48879aebd9576bfb5f68 (Merge pull request #376 from jdleesmiller/fix-cve-2018-1000544) for version 1.2.2, which fixes the Zip Slip (CVE-2018-1000544) vulnerability.

Would it be possible to just delete those files?

Kind regards,
-- Maxime DERCHE

jdleesmiller:
Thanks for reporting this, and sorry for the trouble caused by my patch!

One solution would be to exclude the test files from the gem. I'm not sure what the best practice here is. The discussion here suggests that it would be OK to exclude the test files: https://stackoverflow.com/questions/37800233/should-one-include-tests-in-a-packaged-gem

There are some counterpoints here: https://github.com/rubygems/rubygems/issues/735

A possible compromise would be to exclude the path traversal test files from the gem and adjust the test suite to skip those tests if they are missing.

I'm certainly open to other suggestions.

GElkayam:
hi @jdleesmiller,
Joining in on the request to exclude test_files from gemspec. as can be seen on https://github.com/rubygems/rubygems/issues/735 many packages are removing this as it is not needed anymore.
[...]

PR #405, opened 2019-09-18
Remove test files from gem

jdleesmiller:
Remove the `test_files` from the gemspec, because rubyzip's test files are quite large now and include some malicious zip files that trigger virus scanners on machines that have the gem installed.

Fix https://github.com/rubyzip/rubyzip/issues/384 .

The trend in https://github.com/rubygems/rubygems/issues/735 seems to be to remove `test_files` files from the gemspec. There is some good discussion in that ticket about why this may not be the best thing to do long term (pending changes in rubygems to better support test files), but it does seem to be the preferred approach at the moment.

I am planning to add this to the upcoming 2.0 release. I will leave this open for approximately one week for comment.

CC @MaximeDerche @GElkayam


hainesr:
I think this is a sensible change.

Issue #404, opened 2019-09-16
Update minimum required ruby version to more closely match maintained rubies?


hainesr:
In the gemspec we currently have: `required_ruby_version = '>= 1.9.2'`. The latest maintained ruby is 2.4.

Should we bump up the version of ruby in the gemspec? If not all the way to 2.4 (a bit harsh). maybe at least to 2.2 or (my preference) 2.3?

Requiring a more modern ruby would make things like testing and linting easier, and it would allow us to use nice (and more performant) features as well.

Thoughts (especially @jdleesmiller)?

jdleesmiller:
Yes, I think if we are going to do a major version bump, we should take the opportunity to drop support for EOL rubies. I would be OK with dropping (official) 2.3 support now. It has been EOL for [nearly 6 months](https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/).

PR #406, opened 2019-09-18
Bump supported ruby versions

jdleesmiller:
Upgrade minimum supported ruby versions to include only supported rubies. This is planned as part of an upcoming major release (2.0).

This will hopefully make CI less fragile, allow us to upgrade our linting system, and allow us to take advantage of newer ruby features.

Fixes https://github.com/rubyzip/rubyzip/issues/404 .

- Ruby 2.3 has been EOL since the [end of March 2019](https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/).
- Support for JRuby 9.1 ended [at the end of 2017](https://github.com/jruby/jruby/wiki/Roadmap).
- JDK7 is still in [extended support](https://en.wikipedia.org/wiki/Java_version_history), but it seems that JRuby has [dropped it from CI](https://github.com/jruby/jruby/blob/37e63e4d57354daa419af9abc26a9b14b8010001/.travis.yml#L29-L30), and it has caused problems recently (#399).
- It would be nice to move from `trusty` to `xenial`, but that currently prevents testing on rbx (https://github.com/rubyzip/rubyzip/pull/399). The [canary rbx-4 build](https://travis-ci.org/rubinius/travis-canary/builds/558292567) still seems to be broken. To be fair, the tests fail on rbx, and we don't fail the build, so we could give up on this and move to trusty now, but so far it doesn't seem to cause other problems.

Users who need older versions of ruby will still be able to use the planned 1.3 release for now.

I have tried to update the README and gemspec to make this clearer (and will update the changelog as part of the next release), to avoid confusion like in #254 .

Edit: I will leave this open for approximately one week for comment.

CC @hainesr @simonoff

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,22 +2,16 @@ language: ruby
 dist: trusty
 cache: bundler
 rvm:
-  - 2.0
-  - 2.1
-  - 2.2
-  - 2.3
   - 2.4
   - 2.5
   - 2.6
   - ruby-head
 matrix:
   include:
-  - rvm: jruby
-    jdk: oraclejdk8
-  - rvm: jruby-9.1
-    jdk: openjdk7
+  - rvm: jruby-9.2
+    jdk: openjdk8
   - rvm: jruby-head
-    jdk: oraclejdk8
+    jdk: openjdk8
   - rvm: rbx-4
   allow_failures:
     - rvm: ruby-head
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ gem 'zip-zip' # will load compatibility for old rubyzip API.
 
 ## Requirements
 
-- Ruby 1.9.2 or greater
+- Ruby 2.4 or greater (for rubyzip 2.0; use 1.x for older rubies)
 
 ## Installation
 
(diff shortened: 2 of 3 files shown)


hainesr:
Could we leave a couple of EOL rubies in the "allow failures" section of the CI?

Also, this PR will bump the rubyzip version to 2.0.0, per Semantic Versioning, so should this be included too? Or is it OK to do separately?

Otherwise, great.

jdleesmiller:
Thanks for the feedback!

> Could we leave a couple of EOL rubies in the "allow failures" section of the CI?

I like that idea, but it [seems](https://travis-ci.org/rubyzip/rubyzip/jobs/587623956) that setting the 2.4 version constraint in the gemspec prevents it from running on older rubies:
```
    rubyzip was resolved to 1.2.4, which depends on
      ruby (>= 2.4)
```

I don't know of any straightforward way around that.

> Or is it OK to do separately?

I'm planning to bump the version number separately, after merging the outstanding PRs. I think that will be OK.

hainesr:
Good point about the versions - I obviously had my brain turned off at that point!

Yes, the only solution is to not test the older ones. That's fine. 👍

PR #411, opened 2019-10-01
Clean up temp file usage and fix #410

### LINKED ISSUE #410: Tempfile problems on Windows
The same code used on Windows and Linux behaves differently.
Everything works fine on Linux, but on Windows I get errors when trying to open a stream from a zip file that includes a folder/directory name.

        10: from P:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/rubyzip-2.0.0/lib/zip/filesystem.rb:246:in `open'

         9: from P:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/rubyzip-2.0.0/lib/zip/filesystem.rb:573:in `get_output_stream'

         8: from P:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/rubyzip-2.0.0/lib/zip/file.rb:265:in `get_output_stream'

         7: from P:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/rubyzip-2.0.0/lib/zip/file.rb:265:in `new'

         6: from P:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/rubyzip-2.0.0/lib/zip/streamable_stream.rb:10:in `initialize'

         5: from P:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/rubyzip-2.0.0/lib/zip/streamable_stream.rb:10:in `new'

         4: from P:/Ruby26-x64/lib/ruby/2.6.0/tempfile.rb:131:in `initialize'

         3: from P:/Ruby26-x64/lib/ruby/2.6.0/tmpdir.rb:135:in `create'

         2: from P:/Ruby26-x64/lib/ruby/2.6.0/tempfile.rb:133:in `block in initialize'

         1: from P:/Ruby26-x64/lib/ruby/2.6.0/tempfile.rb:133:in `open'

P:/Ruby26-x64/lib/ruby/2.6.0/tempfile.rb:133:in `initialize': No such file or directory @ rb_sysopen - dvlp/proxy.local__PROJECT$h595f5a46$h4bbd$h4126$h86a6$hec01f155cb67_resource_com$dnomagic$dmagicdraw$duml_umodel$dshared_umodel$dsnapshot20190927-27860-1nww0sq (Errno::ENOENT)

If I alter line 6 in streamable_stream.rb to:
                  nil ##::File.dirname(zipfile)
Which in effect enables Tempfile to use the Dir.tempdir for temporary file location everything seems to work fine.

I would like to understand why the temporary file for Streamable objects is being placed it a local folder instead of just using the Dir.tempdir folder.

hainesr:
There didn't seem to be any particular reason why we weren't using the OS temp directory to store temporary files, and this was reported as an issue on Windows in #410.

This PR fixes that, and also cleans up the usage of `tmpdir` - which isn't needed in the library code, but is used in the tests.


samsonjs:
@jdleesmiller @hainesr This is a problem for me as I have a setup where the root volume is small and there's a second, larger volume dedicated for zipping files. If I were to submit a PR that adds some kind of flag or option to control whether the zip file is created in the system temp directory or the zip directory is it likely to be accepted? (defaulting to the new behaviour)

hainesr:
I think this would be a reasonable option to have. Is it sensible to have the new behaviour the default though?

samsonjs:
Ok great! I'll work on it and should have a PR for you fairly soon.

Now that it's changed I feel like it makes sense to keep it this way. It seems like good behaviour because if zip operation is interrupted for some reason any leftover files are more likely to be cleaned up from /tmp automatically.

I'm happy to code it up however you like though. For our use-case it won't make an appreciable difference either way.

PR #412, opened 2019-10-05
Stable Ruby

GrimmDaddy:
- Create temporary files in the system temporary directory instead of the directory of the zip file [#411](https://github.com/rubyzip/rubyzip/pull/411)
- Drop unused `tmpdir` requirement [#411](https://github.com/rubyzip/rubyzip/pull/411)


hainesr:
This looks like a rogue PR to me!

Issue #395, opened 2019-05-10
Options not working in File class


augustocravosilva:
This code
https://github.com/rubyzip/rubyzip/blob/249775f5637e6d65112574b3ac1763dc2393c7f6/lib/zip/file.rb#L88-L89
doesn't do what it is intended to do, because when you pass in `restore_permissions: false` or `restore_times: false`, it will be evaluating to `false || true`, which will always be `true`

Amorn-Kingthong:
i will go

hainesr:
Interestingly the options code doesn't agree with the comments about defaults above either:

https://github.com/rubyzip/rubyzip/blob/94b7fa276992933592d69eb6bb17fc09105f8395/lib/zip/file.rb#L56-L61

vs

https://github.com/rubyzip/rubyzip/blob/94b7fa276992933592d69eb6bb17fc09105f8395/lib/zip/file.rb#L101-L103

I'll cook up a PR that fixes this issue and works as expected to keep the test passing - adjusting the comments as necessary.

hainesr:
Also, depressingly, it doesn't matter what those options are set to; the tests always all pass. I'll add some tests.

hainesr:
So, on further investigation, I don't think that `restore_times: true` can ever work, because I don't think that when you extract a file it ever even attempts to set the timestamp.

As far as I can tell, this is the code which does the actual extraction to a file:
https://github.com/rubyzip/rubyzip/blob/34d2074ecbf6ef8557c29efc03a585c309be5c4b/lib/zip/entry.rb#L597-L624
And there's nothing there that would restore times...

I'm looking at this in more depth.

jdleesmiller:
Sorry, overzealous auto-closing. This is still a problem.

Issue #398, opened 2019-06-14
Add ability to not store extra attributes


TikiTDO:
I have a scenario where I would like to create a zip that resolves to the same hash each time, similar to how the *nix `zip -X` flag works, but ideally more cross-platform. To that effect it would be nice to have an option to disable the storage of additional fields.

jdleesmiller:
Does this (not yet merged) example help? https://github.com/rubyzip/rubyzip/pull/365

TikiTDO:
That's pretty similar to the snippet I had above, but my main worry with such an approach is that it depends on internal APIs that may end up changing at some point in the future. Ideally this could be done as an option something like `Zip::File.open(name, Zip::File::CREATE, false, reproducible: true)`, with a test case.

I can make a PR to make the `Zip::File#add` function respect an option like that using either my example code above, or the example from the PR you referenced, but I don't want to assume your preferred implementation.

GrimmDaddy:
Zip::File.open(name, Zip::File::CREATE, false, reproducible: true), with a test case.

GrimmDaddy:
Zip::File.open("a.zip", Zip::File::CREATE) do |zipfile|
  a = zipfile.add("a", "a")
  a.time = ::Zip::DOSTime.new(2000)
  a.extra = ::Zip::ExtraField.new
  puts a.class

hainesr:
In terms of locking down the timestamp of a file, we're going through making sure this works correctly in #413, which may actually do what you need here once it's complete.

hainesr:
I do think that the ability to not store extra attributes by default is a good feature to consider. I'm going to look into sensible strategies for implementing this.

I'd be interested in people's thoughts on whether this aught to be a setting applied at the archive, or individual file level, or some combination of both.

jdleesmiller:
Thanks for the PRs, which are both now landed. I am wondering whether we are at a good place for a new release, probably 2.1.0, or whether to hold a bit longer and try to get this in, given that you've been working on it @hainesr ?

> depends on internal APIs

For better or (more likely) for worse, I am not sure there is a very clear distinction between internal and external APIs in this library, at present. So, if we did change the signature of `Zip::Entry.new`, I think we'd have to call it a major version bump in any case.

> whether this ought to be a setting applied at the archive, or individual file level, or some combination of both.

I think there are several possible implementations of this idea:
- Use `mtime` for each file, as suggested in #365
[...]

PR #413, opened 2019-10-07
Fix `restore_times` option when extracting, and test file options

hainesr:
This PR was prompted by #395. When investigating that issue it quickly became obvious that there were deeper problems - as can be seen in the notes for that ticket.

This PR:
* Correctly applies the options specified in ::Zip::File.new, and supplies defaults for any not given.
* Fixes the `restore_times` option by actually storing the original timestamp of a file, and then optionally restoring it on extraction. It also adds tests for this functionality.
* Adds tests for the `restore_permissions` option.
* Does *not* add tests for the `restore_ownership` option. This needs some thinking about because restoring ownership would require other users to be set up on the test machine, and would also potentially require such tests to be run as root.

They are fairly simple fixes, once one has one's head around what is going on in `entry.rb`, but it would be good to have another pair of eyes look at this just in case, @jdleesmiller.

### DIFF
--- a/lib/zip/entry.rb
+++ b/lib/zip/entry.rb
@@ -406,9 +406,7 @@ module Zip
       @unix_uid   = stat.uid
       @unix_gid   = stat.gid
       @unix_perms = stat.mode & 0o7777
-
-      mtime = stat.mtime
-      @time = ::Zip::DOSTime.local(mtime.year, mtime.month, mtime.day, mtime.hour, mtime.min, mtime.sec)
+      @time = ::Zip::DOSTime.from_time(stat.mtime)
     end
 
     def set_unix_attributes_on_path(dest_path)
--- a/lib/zip/entry.rb
+++ b/lib/zip/entry.rb
@@ -406,16 +406,22 @@ module Zip
       @unix_uid   = stat.uid
       @unix_gid   = stat.gid
       @unix_perms = stat.mode & 0o7777
+
+      mtime = stat.mtime
+      @time = ::Zip::DOSTime.local(mtime.year, mtime.month, mtime.day, mtime.hour, mtime.min, mtime.sec)
     end
 
-    def set_unix_permissions_on_path(dest_path)
-      # BUG: does not update timestamps into account
+    def set_unix_attributes_on_path(dest_path)
       # ignore setuid/setgid bits by default.  honor if @restore_ownership
       unix_perms_mask = 0o1777
       unix_perms_mask = 0o7777 if @restore_ownership
       ::FileUtils.chmod(@unix_perms & unix_perms_mask, dest_path) if @restore_permissions && @unix_perms
       ::FileUtils.chown(@unix_uid, @unix_gid, dest_path) if @restore_ownership && @unix_uid && @unix_gid && ::Process.egid == 0
-      # File::utimes()
+
+      # Restore the timestamp on a file. This will either have come from the
+      # original source file that was copied into the archive, or from the
+      # creation date of the archive if there was no original source file.
+      ::FileUtils.touch(dest_path, mtime: time) if @restore_times
     end
 
     def set_extra_attributes_on_path(dest_path) # :nodoc:
@@ -423,7 +429,7 @@ module Zip
 
       case @fstype
       when ::Zip::FSTYPE_UNIX
-        set_unix_permissions_on_path(dest_path)
+        set_unix_attributes_on_path(dest_path)
       end
     end
 
@@ -601,8 +607,6 @@ module Zip
       end
       ::File.open(dest_path, 'wb') do |os|
         get_input_stream do |is|
-          set_extra_attributes_on_path(dest_path)
-
           bytes_written = 0
           warned = false
           buf = ''.dup
@@ -621,6 +625,8 @@ module Zip
           end
         end
       end
+
+      set_extra_attributes_on_path(dest_path)
     end
 
     def create_directory(dest_path)
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -59,12 +59,15 @@ module Zip
 
     attr_reader :name
 
-    # default -> false
+    # default -> false.
     attr_accessor :restore_ownership
-    # default -> false
+
+    # default -> false, but will be set to true in a future version.
     attr_accessor :restore_permissions
-    # default -> true
+
+    # default -> false, but will be set to true in a future version.
     attr_accessor :restore_times
+
     # Returns the zip files comment, if it has one
     attr_accessor :comment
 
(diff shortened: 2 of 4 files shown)


lib/zip/file.rb
```diff
@@ -51,21 +51,31 @@ class File < CentralDirectory
     DATA_BUFFER_SIZE     = 8192
     IO_METHODS           = [:tell, :seek, :read, :close]
 
+    DEFAULT_OPTIONS = {
+      restore_ownership:   false,
+      restore_permissions: false,
+      restore_times:       true
```
jdleesmiller: While the default value for `restore_times` was notionally `true` in the original code, it didn't actually do anything (as you discovered), so it effectively defaulted to `false`. For compatibility, perhaps this should default `false` for now?

lib/zip/file.rb
```diff
@@ -51,21 +51,31 @@ class File < CentralDirectory
     DATA_BUFFER_SIZE     = 8192
     IO_METHODS           = [:tell, :seek, :read, :close]
 
+    DEFAULT_OPTIONS = {
+      restore_ownership:   false,
+      restore_permissions: false,
```
jdleesmiller: It looks like this used to default to `true` (and, unlike `restore_times`, did actually do something). Is there a particular reason to default it to `false` now?

lib/zip/entry.rb
```diff
@@ -406,24 +406,28 @@ def get_extra_attributes_from_path(path) # :nodoc:
       @unix_uid   = stat.uid
       @unix_gid   = stat.gid
       @unix_perms = stat.mode & 0o7777
+      @time = ::Zip::DOSTime.from_time(stat.mtime)
```
jdleesmiller: - This is currently guarded by `return if Zip::RUNNING_ON_WINDOWS`. I am pretty sure that the modification time would also be available on Windows, and from what I can see in the documentation, `File::stat` should still work. I have been meaning to find a Windows machine to test this on... I guess even if the `restore_times` only works on Linux, it would still be an improvement.

- What do you think to using `time=` rather than assigning the `@time` instance variable directly? It looks like that would also set some related attributes in `@extra`, which seems desirable.

lib/zip/entry.rb
```diff
@@ -406,24 +406,28 @@ def get_extra_attributes_from_path(path) # :nodoc:
...
       unix_perms_mask = 0o1777
       unix_perms_mask = 0o7777 if @restore_ownership
       ::FileUtils.chmod(@unix_perms & unix_perms_mask, dest_path) if @restore_permissions && @unix_perms
       ::FileUtils.chown(@unix_uid, @unix_gid, dest_path) if @restore_ownership && @unix_uid && @unix_gid && ::Process.egid == 0
-      # File::utimes()
+
+      # Restore the timestamp on a file. This will either have come from the
+      # original source file that was copied into the archive, or from the
+      # creation date of the archive if there was no original source file.
+      ::FileUtils.touch(dest_path, mtime: time) if @restore_times
```
jdleesmiller: Possibly one for a future PR... but we could use `File::utime` here as the original comment hinted, if we used the `atime` and `ctime` from the [`UniversalTime`](https://github.com/rubyzip/rubyzip/blob/8887b703490337f7ab68ebfb4880354db28ba648/lib/zip/extra_field/universal_time.rb#L10) extra field. (And we could store those from `File::stat`, too.)

lib/zip/file.rb
hainesr: I think that restoring the original times probably should be the eventual default - I've seen use-cases where the whole point of zipping a directory up and transferring it was to preserve the timestamps.

But I take your point that it might be better to default to false for now. I'll change that.

lib/zip/file.rb
hainesr: It did default to `true`, but it didn't do anything! I forgot to mention this in the original PR.

In `Entry#create_file` the call to set the attributes of a file (`set_extra_attributes_on_path`) was being called too early - before the file was closed - so the permissions weren't being set correctly. In the course of my PR I moved the call to `set_extra_attributes_on_path` to when the file is closed to fix this. (`Entry#create_directory` got this right already.)

Then I set this to be be `false` by default for consistency. I certainly think that permissions *should* be restored by default though.

I'm happy to change this to `true` while I'm making the other changes required.

lib/zip/entry.rb
hainesr: Good point, I will try using `time=`. TBH working through how all this works was horribly convoluted so I must have just missed this subtlety 😬

lib/zip/file.rb
jdleesmiller: Aha, I didn't pick up on that. 👍 for `false` in that case.

lib/zip/entry.rb
hainesr: I think I used `mtime` because that is the only one `FileUtils.touch` supports. But now I look at the code for `FileUtils.touch` it turns out it uses `File.utime` internally anyway [1], setting `atime` and `mtime` to the same thing.

So, I agree, shall we leave this as is for now and look at it in a new PR?

[1] https://ruby-doc.org/stdlib-2.4.7/libdoc/fileutils/rdoc/FileUtils.html#method-c-touch

lib/zip/file.rb
hainesr: OK, I will leave it set to false for now then.

lib/zip/entry.rb
jdleesmiller: Good points. I am definitely OK with that 👍.

lib/zip/entry.rb
hainesr: Urgh. So I don't think the `UniversalTime` extra field can work as currently implemented. In `Entry#time=` when it creates a `UniversalTime` entry in `@extra` the `.flag` field is set to `nil`. But the `UniversalTime` field can only be packed if `@flag` is non-`nil`.

Going through the tests for the extra field code it seems that `@flag` is always set to 3 for `UniversalTime`, and indeed forcing this does work. But I need to look and see if this reasonable.

It's late now, so I'll pick this up another time.

hainesr: Right. I think given the issues with using `time=` at the moment, I think it's best to keep using `@time` for now.

I've found documentation for the `UniversalTime` extra field now, so I think I can fix it in a different PR once I've properly got my head round it.

I also think, now I've seen what adding the extra fields does to the archive, that having a way of *not* using extra fields by default (as suggested in #398) would be a good idea. So I'll have a think about that too.

I'll finish up making the other changes to this PR we've already discussed this evening and update it.

hainesr:
I have made the changes we discussed above - apart from the switch to using `time=` for now as explained - and rebased everything onto the tip of master.

PR #420, opened 2019-10-30
Quickly fire up a console with 'zip' pre-loaded.

hainesr:
I find myself testing things in `irb` enough that this has proved useful, so I decided to commit it.

It's purely a developer nicety so isn't stored in the gem.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -1,0 +1,7 @@
+#!/usr/bin/env ruby
+
+require 'bundler/setup'
+require 'zip'
+
+require 'irb'
+IRB.start(__FILE__)


bin/console
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env ruby
+
+lib = File.expand_path(File.join('..', 'lib'), __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'zip'
```
jdleesmiller: I think it's possible to do this with `require_relative` now (I think since 1.9), without the path manipulation:
```rb
require_relative '../lib/zip'
```

hainesr: I always forget about `require_relative`. I'll change that.

hainesr: Having now tried this, `require_relative` doesn't actually work, because everything in `lib/zip.rb` is `require`d, so also relies on `$LOAD_PATH` being set correctly.

So I could fix this, but it would be a lot more involved that just changing this file. Happy to do so if you think it's worth it?

hainesr:
The problem with `exec irb -I lib -r zip` is that it doesn't take `cwd` into account. To be able to run both
* `./bin/console`
* `cd bin; ./console`

you'd need something like `exec irb -I lib:../lib -r zip`, and then it still wouldn't be that robust. I'm tempted to prefer the more wordy ruby version.

jdleesmiller:
OK, the current script isn't bad, and I don't want to belabour the point, but to make sure we've gone through the options:

- I tried it with `require_relative '../lib/zip'` and the resulting console seemed to work --- what didn't work when you tried it?
- That is true about the shell script assuming it's run from the root. I have never found a particularly elegant solution for this problem, but my usual approach has been to rely on `bash` with something like
```sh
#!/usr/bin/env bash
exec irb -I $(dirname "${BASH_SOURCE[0]}")../lib -r zip
```

hainesr:
Sorry, forgot about this one for a while. The other reason I went the ruby (rather that shell) route, is that this mimics the current working of bundler: when you're starting a gem from scratch you can do `bundler gem name` and get a load of nice boilerplate created for you.

I've now gone back and looked at how bundler sets this file up and adjusted it accordingly. I think it's pretty neat now, and doesn't rely on shell variables and so on.

PR #421, opened 2019-10-30
Fixes for the UniversalTime extra field

### LINKED ISSUE #395: Options not working in File class

hainesr:
When working on #413 to fix #395 I discovered that there were a few problems with how the timestamps were being stored - particularly with the UniversalTime extra field. On even closer inspection, and after having consulted the UniversalTime specification (ftp://ftp.info-zip.org/pub/infozip/doc/), I found further issues.

This PR fixes those issues and makes the `UniversalTime` class more user-friendly. It also adds tests.

At a later date I will fold this in so that it is used to store and restore the timestamps of files, but I've left this for now as it has the side-effect of adding data (that might be unexpected/unwanted) between the filename and file data. In short, I wonder if we should allow some control over which extra fields are used, rather than inserting them by default, c.f. #398.

### DIFF
--- a/lib/zip/extra_field/universal_time.rb
+++ b/lib/zip/extra_field/universal_time.rb
@@ -13,7 +13,8 @@ module Zip
       @mtime = nil
       @atime = nil
       @flag  = 0
-      binstr && merge(binstr)
+
+      merge(binstr) unless binstr.nil?
     end
 
     attr_reader :atime, :ctime, :mtime, :flag
@@ -58,15 +59,15 @@ module Zip
 
     def pack_for_local
       s = [@flag].pack('C')
-      @flag & 1 != 0 && s << [@mtime.to_i].pack('l<')
-      @flag & 2 != 0 && s << [@atime.to_i].pack('l<')
-      @flag & 4 != 0 && s << [@ctime.to_i].pack('l<')
+      s << [@mtime.to_i].pack('l<') unless @flag & MTIME_MASK == 0
+      s << [@atime.to_i].pack('l<') unless @flag & ATIME_MASK == 0
+      s << [@ctime.to_i].pack('l<') unless @flag & CTIME_MASK == 0
       s
     end
 
     def pack_for_c_dir
       s = [@flag].pack('C')
-      @flag & 1 == 1 && s << [@mtime.to_i].pack('l<')
+      s << [@mtime.to_i].pack('l<') unless @flag & MTIME_MASK == 0
       s
     end
   end
--- a/lib/zip/extra_field/universal_time.rb
+++ b/lib/zip/extra_field/universal_time.rb
@@ -4,24 +4,43 @@ module Zip
     HEADER_ID = 'UT'
     register_map
 
+    ATIME_MASK = 0b010
+    CTIME_MASK = 0b100
+    MTIME_MASK = 0b001
+
     def initialize(binstr = nil)
       @ctime = nil
       @mtime = nil
       @atime = nil
-      @flag  = nil
+      @flag  = 0
       binstr && merge(binstr)
     end
 
-    attr_accessor :atime, :ctime, :mtime, :flag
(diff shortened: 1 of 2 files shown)


hainesr:
Sorry to bump this, but would you mind having a look @jdleesmiller?

lib/zip/extra_field/universal_time.rb
```diff
@@ -8,20 +8,52 @@ def initialize(binstr = nil)
...
-      binstr && merge(binstr)
+      @flag  = 0
+
+      merge(binstr) unless binstr.nil?
+    end
+
+    attr_reader :atime, :ctime, :mtime, :flag
+
+    def atime=(time)
+      time.nil? ? @flag &= 0b101 : @flag |= 0b010
```
jdleesmiller: Maybe it's worth extracting these masks into constants? Something like
```rb
ATIME_MASK = 0b010
@flag = time.nil? ? @flag & ~ATIME_MASK : @flag | ATIME_MASK
```

lib/zip/extra_field/universal_time.rb
```diff
@@ -8,20 +8,52 @@ def initialize(binstr = nil)
...
-      mtime && @mtime ||= ::Zip::DOSTime.at(mtime)
-      atime && @atime ||= ::Zip::DOSTime.at(atime)
-      ctime && @ctime ||= ::Zip::DOSTime.at(ctime)
+      return if size.nil? || size <= 0
+
+      flags, *times = content.unpack('Cl<l<l<')
+      @flag = flags
+
+      # Parse the timestamps, in order, based on which flag is set.
+      times.each do |time|
```
jdleesmiller: I think your reading of the spec is correct --- good find. I think it may be (marginally) clearer without the loop, if this also looks right to you?
```rb
@flag, *times = content.unpack('Cl<3')
return if times[0].nil?
@mtime ||= ::Zip::DOSTime.at(times.shift) unless @flag & MTIME_MASK == 0
return if times[0].nil?
@atime ||= ::Zip::DOSTime.at(times.shift) unless @flag & ATIME_MASK == 0
return if times[0].nil?
@ctime ||= ::Zip::DOSTime.at(times.shift) unless @flag & CTIME_MASK == 0
```

lib/zip/extra_field/universal_time.rb
hainesr: Agreed. I'll do that.

lib/zip/extra_field/universal_time.rb
```diff
@@ -8,20 +8,52 @@ def initialize(binstr = nil)
...
     def merge(binstr)
       return if binstr.empty?
+
       size, content = initial_parse(binstr)
-      size || return
-      @flag, mtime, atime, ctime = content.unpack('CVVV')
-      mtime && @mtime ||= ::Zip::DOSTime.at(mtime)
-      atime && @atime ||= ::Zip::DOSTime.at(atime)
-      ctime && @ctime ||= ::Zip::DOSTime.at(ctime)
+      return if size.nil? || size <= 0
```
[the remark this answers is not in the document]
hainesr: Hmm, yes. That should be `!size` not `size.nil?`. Will fix.

lib/zip/extra_field/universal_time.rb
hainesr: This is somewhat neater and doesn't need a second copy of the flags. I'll go with this.

hainesr:
Thanks for the review. I have made the above fixes and have rebased onto current `master`.

PR #425, opened 2019-11-29
Bzip2

jspanjers:


hainesr:
Can I ask why you copied in the bz2 ffi stuff rather than stayed using the dependency?

jspanjers:
> Can I ask why you copied in the bz2 ffi stuff rather than stayed using the dependency?

I started development depending on bzip2-ffi.

Afterwards I realized that I was using the private internal api from bzip2-ffi (Bzip2::FFI::Libbz2). I think it is not right to make rubyzip depend on non public api from bzip2-ffi.

I noticed that I was only using the Libbz2 ffi bindings, which are just 75 lines of code. Copying these bindings to rubyzip prevents the use of private api and also removes all dependencies on bzip2-ffi, leaving only a dependency on ffi.

hainesr:
+1 I like/prefer the idea of a separate `rubyzip-bzip2` gem for this.

jdleesmiller:
Closing in view of #427.

PR #432, opened 2020-02-06
Add an option to create temp files in the zip file directory

samsonjs:
Before zip files were created in the system temporary directory they
were created in the same directory as the zip file. This adds an option
to have the same behaviour as before by passing in
`use_system_temp_dir: false` as an option to `Zip.open`, which passes
that down to `Zip::StreamableStream`.

See #411 for discussion leading up to this PR.

I didn't see an obvious way to test that `Zip::File` passes the option down to `Zip::StreamableStream` but am happy to add such a test with a bit of direction from the core maintainers.

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -54,7 +54,8 @@ module Zip
     DEFAULT_OPTIONS = {
       restore_ownership:   false,
       restore_permissions: false,
-      restore_times:       false
+      restore_times:       false,
+      use_system_temp_dir: true
     }.freeze
 
     attr_reader :name
@@ -111,6 +112,7 @@ module Zip
       @restore_ownership   = options[:restore_ownership]
       @restore_permissions = options[:restore_permissions]
       @restore_times       = options[:restore_times]
+      @use_system_temp_dir = options[:use_system_temp_dir]
     end
 
     class << self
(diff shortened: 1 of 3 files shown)


hainesr:
Just as a sanity check: are we sure that this option should be in `File` and not at the top level?

jdleesmiller:
Thanks for the PR, and sorry to hear that this change caused a problem.

I think there are a few other options here:

- You can [put `/tmp` on another partition](https://askubuntu.com/questions/371619/move-tmp-folder-from-partition-to-mounted-partition-mnt). I think that is a fairly common approach when large a large scratch space is required.

- Ruby's [`Dir::tmpdir`](https://ruby-doc.org/stdlib-2.7.0/libdoc/tmpdir/rdoc/Dir.html#method-c-tmpdir) respects the `TMPDIR` environment variable, so you could point ruby's `tmpdir` at a location on your larger volume:
   ```
   # on mac with ruby 2.6.3
   $ ruby -r tempfile -e 'puts Tempfile.new.path'
   /var/folders/xq/6tgxlyhx2yz8qk1mq575qg1m0000gn/T/20200208-46464-1vib46p
   $ TMPDIR=. ruby -r tempfile -e 'puts Tempfile.new.path'
   /Users/john/20200208-46480-oal377
   ```

Would one of those help?

samsonjs:
Thanks a lot for the suggestions @jdleesmiller. We just deployed the `TMPDIR` solution to our staging environment and it's working great, so I'll close this PR. Really appreciate you taking the time to help us out.

PR #439, opened 2020-02-19
Set OutputStream.write_buffer's default buffer to binmode

### LINKED ISSUE #438: OutputStream.write_buffer returns IOString with default encoding
In my case this is UTF-8, which produces encoding errors in encoding sensitive use cases.

I'd rather initialize the StringIO.new() in binmode which sets the encoding to ASCI-8BIT and prevents such errors, which are hard to find.

henkesn:
Fixes #438

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -362,7 +362,6 @@ module Zip
 
     # Write buffer write changes to buffer and return
     def write_buffer(io = ::StringIO.new(''))
-      io.binmode if io.respond_to?(:binmode)
       ::Zip::OutputStream.write_buffer(io) do |zos|
         @entry_set.each { |e| e.write_to_zip_output_stream(zos) }
         zos.comment = comment
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -362,6 +362,7 @@ module Zip
 
     # Write buffer write changes to buffer and return
     def write_buffer(io = ::StringIO.new(''))
+      io.binmode if io.respond_to?(:binmode)
       ::Zip::OutputStream.write_buffer(io) do |zos|
         @entry_set.each { |e| e.write_to_zip_output_stream(zos) }
         zos.comment = comment
--- a/lib/zip/output_stream.rb
+++ b/lib/zip/output_stream.rb
@@ -58,6 +58,7 @@ module Zip
 
       # Same as #open but writes to a filestream instead
       def write_buffer(io = ::StringIO.new(''), encrypter = nil)
+        io.binmode if io.respond_to?(:binmode)
         zos = new(io, true, encrypter)
         yield zos
         zos.close_buffer
--- a/test/output_stream_test.rb
+++ b/test/output_stream_test.rb
@@ -6,6 +6,8 @@ class ZipOutputStreamTest < MiniTest::Test
   TEST_ZIP = TestZipFile::TEST_ZIP2.clone
   TEST_ZIP.zip_name = 'test/data/generated/output.zip'
 
+  ASCII8BIT = 'ASCII-8BIT'
+
   def test_new
     zos = ::Zip::OutputStream.new(TEST_ZIP.zip_name)
     zos.comment = TEST_ZIP.comment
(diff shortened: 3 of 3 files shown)


lib/zip/output_stream.rb
```diff
@@ -57,7 +57,7 @@ def open(file_name, encrypter = nil)
       end
 
       # Same as #open but writes to a filestream instead
-      def write_buffer(io = ::StringIO.new(''), encrypter = nil)
+      def write_buffer(io = ::StringIO.new('').binmode, encrypter = nil)
```
hainesr: What if we pass an `IO` in here that hasn't been set to `binmode`? Do we still see the behaviour you're trying to avoid? This change only sets a newly created `StringIO` to `binmode`. Is this a problem?

How about:
```suggestion
      def write_buffer(io = ::StringIO.new(''), encrypter = nil)
        io.binmode if io.respond_to?(:binmode)
```
Which would match behaviour in `::Zip::File.open_buffer`?

henkesn: I found another location where the binmode would be set if supported. I think this is a cleaner solution:
https://github.com/henkeinfo/rubyzip/blob/5ce4e13ddd33443f01fa33c8208423b76d99fce3/lib/zip/file.rb#L151
Yes, you're right with your comment. I rewrote the `write_buffer` to behave like the source above and also fixed the same thing in yet another method.

hainesr: Yes I spotted that when I was looking through `File` and wondered why it was inconsistent. Thanks for the catch.

This looks good to me.

lib/zip/file.rb
```diff
@@ -362,6 +362,7 @@ def commit
 
     # Write buffer write changes to buffer and return
     def write_buffer(io = ::StringIO.new(''))
+      io.binmode if io.respond_to?(:binmode)
```
jdleesmiller: The tests still pass without this line. I think that given that the first thing it does is call `::Zip::OutputStream.write_buffer`, which calls `binmode`, this line can be deleted.

test/output_stream_test.rb
```diff
@@ -32,6 +34,15 @@ def test_write_buffer
     assert_test_zip_contents(TEST_ZIP)
   end
 
+  def test_write_buffer_binmode
+    io = ::StringIO.new('')
+    buffer = ::Zip::OutputStream.write_buffer(io) do |zos|
+      zos.comment = TEST_ZIP.comment
+      write_test_zip(zos)
+    end
+    assert buffer.external_encoding.name === ASCII8BIT
```
jdleesmiller: Ruby has a constant for this encoding already:
```suggestion
    assert_equal Encoding::ASCII_8BIT, buffer.external_encoding
```

lib/zip/file.rb
henkesn: Yes, you're right. I'll remove this line.

henkesn:
Thank you for your feedback @jdleesmiller.  My use case was an extension in my code where i filtered empty strings before storing to a database (`String#presence`).  I already faced some issues with ActiveRecord in the past, when trying to save invalid UTF8 strings even for BLOB columns. But that's already some time ago and I neither know how to reproduce nor if that issues persist with newer versions.
Regarding the next line
https://github.com/rubyzip/rubyzip/blob/8d91d001fd214777f7a5a61ed2a3b919c33f4c8c/lib/zip/file.rb#L150-L151
perhaps this line now is also superfluous because it internally calls `OutputStream#write_buffer` with my fix. But the yield call is done after this line and before the `OutputStream#write_buffer`, so removing this line potentially could break things.

test/output_stream_test.rb
```diff
@@ -6,6 +6,8 @@ class ZipOutputStreamTest < MiniTest::Test
   TEST_ZIP = TestZipFile::TEST_ZIP2.clone
   TEST_ZIP.zip_name = 'test/data/generated/output.zip'
 
+  ASCII8BIT = 'ASCII-8BIT'
```
jdleesmiller: This constant can now be removed, but I'll tidy it up post-merge.

henkesn: Thx!

kitsunde:
This appears to have affected an upstream gem because of the encoding changing. https://github.com/weshatheleopard/rubyXL/issues/377

PR #440, opened 2020-02-20
Rubocop name cops

hainesr:
This PR goes through and cleans up rubocop naming cops that don't affect the API.

There were many variables/parameters/etc named in `camelCase` style, possibly a hangover from the early days of this library and its Java-like roots. This PR renames them to ruby standard `snake_case` style.

Most names have been translated as simply as possible - e.g. `entryName` to `entry_name` - but some have been simplified as well - e.g. `aProc` to `proc`. There are no changes to functionality in this PR; any other changes, such as formatting, have only been done to conform to our existing rubocop standards.

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -319,11 +319,11 @@ module Zip
 
     # Renames the specified entry.
     def rename(entry, new_name, &continue_on_exists_proc)
-      foundEntry = get_entry(entry)
+      found_entry = get_entry(entry)
       check_entry_exists(new_name, continue_on_exists_proc, 'rename')
-      @entry_set.delete(foundEntry)
-      foundEntry.name = new_name
-      @entry_set << foundEntry
+      @entry_set.delete(found_entry)
+      found_entry.name = new_name
+      @entry_set << found_entry
     end
 
     # Replaces the specified entry with the contents of src_path (from
@@ -420,15 +420,15 @@ module Zip
     private
 
     def directory?(new_entry, src_path)
-      srcPathIsDirectory = ::File.directory?(src_path)
-      if new_entry.directory? && !srcPathIsDirectory
+      path_is_directory = ::File.directory?(src_path)
+      if new_entry.directory? && !path_is_directory
         raise ArgumentError,
               "entry name '#{new_entry}' indicates directory entry, but " \
                   "'#{src_path}' is not a directory"
-      elsif !new_entry.directory? && srcPathIsDirectory
+      elsif !new_entry.directory? && path_is_directory
         new_entry.name += '/'
       end
-      new_entry.directory? && srcPathIsDirectory
+      new_entry.directory? && path_is_directory
     end
 
     def check_entry_exists(entry_name, continue_on_exists_proc, proc_name)
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -169,8 +169,8 @@ module Zip
       # local entry headers (which contain the same information as the
       # central directory).
       def foreach(aZipFileName, &block)
-        ::Zip::File.open(aZipFileName) do |zipFile|
-          zipFile.each(&block)
+        ::Zip::File.open(aZipFileName) do |zip_file|
+          zip_file.each(&block)
         end
       end
 
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -168,8 +168,8 @@ module Zip
       # whereas ZipInputStream jumps through the entire archive accessing the
       # local entry headers (which contain the same information as the
       # central directory).
-      def foreach(aZipFileName, &block)
-        ::Zip::File.open(aZipFileName) do |zip_file|
+      def foreach(zip_file_name, &block)
+        ::Zip::File.open(zip_file_name) do |zip_file|
           zip_file.each(&block)
         end
       end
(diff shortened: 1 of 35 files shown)


test/local_entry_test.rb
```diff
@@ -41,56 +41,71 @@ def test_read_local_entry_from_non_zip_file
...
     entry = ::Zip::Entry.new
-    entry.read_local_entry(zipFragment)
+    entry.read_local_entry(fragment)
     raise 'ZipError expected'
   rescue ::Zip::Error
   end
 
   def test_write_entry
-    entry = ::Zip::Entry.new('file.zip', 'entryName', 'my little comment',
+    entry = ::Zip::Entry.new('file.zip', 'entry_name', 'my little comment',
```
jdleesmiller: I think this is fine, but is this change actually required? (There are a few more like it below.)

test/settings_test.rb
```diff
@@ -17,14 +17,14 @@ def teardown
     ::Zip.reset!
   end
 
-  def open_zip(&aProc)
-    refute_nil(aProc)
-    ::Zip::File.open(TestZipFile::TEST_ZIP4.zip_name, &aProc)
+  def open_zip(&proc)
+    refute_nil(proc)
+    ::Zip::File.open(TestZipFile::TEST_ZIP4.zip_name, &proc)
```
jdleesmiller:
```suggestion
  def open_zip(&block)
    assert block_given?
    ::Zip::File.open(TestZipFile::TEST_ZIP4.zip_name, &block)
```

test/settings_test.rb
```diff
@@ -17,14 +17,14 @@ def teardown
...
   end
 
-  def extract_test_dir(&aProc)
+  def extract_test_dir(&proc)
```
jdleesmiller:
```suggestion
  def extract_test_dir(&block)
```

test/settings_test.rb
```diff
@@ -17,14 +17,14 @@ def teardown
...
     open_zip do |zf|
-      zf.extract(TestFiles::EMPTY_TEST_DIR, TEST_OUT_NAME, &aProc)
+      zf.extract(TestFiles::EMPTY_TEST_DIR, TEST_OUT_NAME, &proc)
```
jdleesmiller:
```suggestion
      zf.extract(TestFiles::EMPTY_TEST_DIR, TEST_OUT_NAME, &block)
```

test/filesystem/file_nonmutating_test.rb
```diff
@@ -126,19 +126,19 @@ def test_file?
   include ExtraAssertions
 
   def test_dirname
-    assert_forwarded(File, :dirname, 'retVal', 'a/b/c/d') do
+    assert_forwarded(File, :dirname, 'ret_val', 'a/b/c/d') do
```
jdleesmiller: This is OK, but the tests still pass without this change (and the next two changes). (I am not really sure why this parameter exists for this helper; it just gets compared to itself, AFAICT.)

test/file_test.rb
```diff
@@ -62,26 +62,26 @@ def test_get_input_stream_stored_with_gpflag_bit3
...
 
   def test_get_output_stream
-    entryCount = nil
+    count = nil
     ::Zip::File.open(TEST_ZIP.zip_name) do |zf|
-      entryCount = zf.size
-      zf.get_output_stream('newEntry.txt') do |os|
-        os.write 'Putting stuff in newEntry.txt'
+      count = zf.size
+      zf.get_output_stream('new_entry.txt') do |os|
```
jdleesmiller: I think this change is not actually needed?

test/file_test.rb
```diff
@@ -503,52 +514,58 @@ def test_commit_use_zip_entry
   #  end
 
   def test_compound1
-    renamedName = 'renamedName'
+    renamed_name = 'renamed_name'
```
jdleesmiller: Technically would still work with the old value, I think.

test/file_test.rb
```diff
@@ -560,55 +577,55 @@ def test_compound2
       end
       assert_equal(zf.entries.sort.map(&:name), TestFiles::ASCII_TEST_FILES)
 
-      zf.rename(TestFiles::ASCII_TEST_FILES[0], 'newName')
+      zf.rename(TestFiles::ASCII_TEST_FILES[0], 'new_name')
```
jdleesmiller: Could do without this change?

lib/zip/file.rb
```diff
@@ -255,8 +255,8 @@ def split(zip_file_name, segment_size = MAX_SEGMENT_SIZE, delete_zip_file = true
     # Returns an input stream to the specified entry. If a block is passed
     # the stream object is passed to the block and the stream is automatically
     # closed afterwards just as with ruby's builtin File.open method.
-    def get_input_stream(entry, &aProc)
-      get_entry(entry).get_input_stream(&aProc)
+    def get_input_stream(entry, &proc)
+      get_entry(entry).get_input_stream(&proc)
```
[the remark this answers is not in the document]
hainesr: Good point; I'll change them. I think `a_proc` might be better in terms of semantically signposting what is being passed in, so I'll go with that where appropriate.

test/filesystem/file_nonmutating_test.rb
hainesr: Yes, I realise these are just strings and they don't need to be done for rubocop... But I think they just look neater and more in keeping with everything else once all the other camel case names have been changed.

test/local_entry_test.rb
hainesr: I was just being comprehensive with getting rid of camel case.

test/file_test.rb
hainesr: Yes I could have left the value unchanged, but was being competitionist about things :-)

hainesr:
I've done the `proc` -> `a_proc` changes. You were definitely right about it being potentially dangerous, and I'm surprised that Rubocop doesn't have an opinion on it, like it does for `open`...

hainesr:
Thanks. Point taken about changing code and tests together; I agree. But as you say these changes are fairly safe in this instance...

PR #444, opened 2020-03-14
Add rubocop to travis CI

hainesr:
Update rubocop to a more recent version and add a rubocop job to the travis config.

This means that any changes to rubyzip will fail in travis if they break the rubocop rules.

See https://travis-ci.org/hainesr/rubyzip/jobs/662466051 for an example run.


rubyzip.gemspec
```diff
@@ -25,5 +25,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'minitest', '~> 5.4'
   s.add_development_dependency 'pry', '~> 0.10'
   s.add_development_dependency 'rake', '~> 12.3', '>= 12.3.3'
-  s.add_development_dependency 'rubocop', '~> 0.79'
+  s.add_development_dependency 'rubocop', '0.80'
```
jdleesmiller: I think this version constraint translates to `= 0.80`, but it looks like there are both `0.80.0` and `0.80.1`, and I am not 100% sure which one you get (it seems a bit underspecified). For consistency with the above, how about:
```suggestion
  s.add_development_dependency 'rubocop', '~> 0.80.1'
```
It seems unlikely that there will be more patch versions 0.80.x, but we'd allow them.

hainesr: I locked it down hard because rubocop has a tendency to introduce breaking changes in minor versions. But, OK, yes, let's at least trust them not to do this in patch versions :grimacing:

I'll fix and update.

hainesr: Done

PR #448, opened 2020-05-24
Set compression level on a per-zipfile basis.

### LINKED ISSUE #433: Set compression on a per-operation basis
There may already be an API for this that I'm not seeing, but it would be nice to be able to set the compression on a per-operation basis.  Something along the lines of:

```  def write(compress: false)
    entries = Dir.entries(@inputDir); entries.delete("."); entries.delete("..")
    io = Zip::File.open(
      @outputFile, Zip::File::CREATE,
      { compression: Zlib::BEST_COMPRESSION } # New Part
    );

    writeEntries(entries, "", io)
    io.close();
  end```

Right at the moment, it only appears possible (based on the documentation) to set this setting globally, which is rather annoying.

### PR DESCRIPTION
**Pull request by hainesr**
This is an early attempt to fix #433.

It was previously only possible to set the compression level (e.g. fastest vs smallest) via `Zip.default_compression`, but that is set at the top level and doesn't allow for it to be set per-zipfile. The ability to set per-zipfile compression level was never quite surfaced high-enough up in the API, and this PR fixes that.

Use it like this:

```ruby
::Zip::File.open('test.zip', true, { compression_level: Zlib::BEST_COMPRESSION }) do |zf|
  zf.add('file.txt')
end
```

Considerations:
* This changes the `Entry` API. If this is considered to be a public API then this PR will require a major version bump. I'm not sure that `Entry` should be considered public, but it may be defacto public.
* I've deliberately kept `Entry.compression_level` out of the public API because it makes no sense when reading entries in a zipfile - from what I can tell it's a write flag that isn't needed, or stored, for decompression.
* I've taken the opportunity to clean up some rather odd behaviour in `OutputStream` which would have overwritten settings in an `Entry` if misused. No tests used this behaviour so it seemed safe to do, and it seems sensible to have `Entry` be a single point of truth where possible.

I'd welcome thoughts on this, and suggestions for improvements. I'll add documentation, etc once we're happy it's a good change.


lib/zip/entry.rb
```diff
@@ -577,10 +578,13 @@ def gather_fileinfo_from_srcpath(src_path) # :nodoc:
 
     def write_to_zip_output_stream(zip_output_stream) #:nodoc:all
       if @ftype == :directory
-        zip_output_stream.put_next_entry(self, nil, nil, ::Zip::Entry::STORED)
+        @compression_method = ::Zip::Entry::STORED
```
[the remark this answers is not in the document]
hainesr: Yes, I think I like that idea. I'll add that change.

hainesr:
Thanks for looking at this @jdleesmiller. I did wonder about looking at the `Entry.new` interface, given that we're changing it anyway.

Also, I spotted something else while looking into some of the issues people have been having with `docx` files: the general purpose bit flag (`gp_flags` in `Entry`) are used to indicate something about compression level used. I _think_ it's just cosmetic, but they are honoured by the `zip` and `zipinfo` tools I have on my Ubuntu installation. For example:

**Level 1 compression (fastest)**
```shell
$ zip -1 ctest-1.zip test.tar
$ zipinfo ctest-1.zip
Archive:  ctest-1.zip
Zip file size: 1737203 bytes, number of entries: 1
-rw-rw-r--  3.0 unx  6809600 bx defF 20-May-30 14:30 t.tar
1 file, 6809600 bytes uncompressed, 1737043 bytes compressed:  74.5%
```

**Level 5 compression**
```shell
$ zip -5 ctest-5.zip test.tar
$ zipinfo ctest-5.zip
Archive:  ctest-5.zip
Zip file size: 1441428 bytes, number of entries: 1
-rw-rw-r--  3.0 unx  6809600 bx defN 20-May-30 14:30 t.tar
1 file, 6809600 bytes uncompressed, 1441268 bytes compressed:  78.8%
```

**Level 9 compression (best)**
```shell
$ zip -9 ctest-9.zip test.tar
$ zipinfo ctest-9.zip
Archive:  ctest-9.zip
Zip file size: 1413035 bytes, number of entries: 1
-rw-rw-r--  3.0 unx  6809600 bx defX 20-May-30 14:30 t.tar
1 file, 6809600 bytes uncompressed, 1412875 bytes compressed:  79.3%
```

Note the `defF`, `defN` and `defX` in the `zipinfo` output. If you look at those files with rubyzip and interrogate the `gp_flags` they are set according to this table in the [PKWARE APPNOTE](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT):

```
(For Methods 8 and 9 - Deflating)
Bit 2  Bit 1
0      0    Normal (-en) compression option was used.
0      1    Maximum (-exx/-ex) compression option was used.
1      0    Fast (-ef) compression option was used.
1      1    Super Fast (-es) compression option was used.
```

And I've seen `defS` inside `docx` files. I presume the `defx` mappings are S = super fast, F = fast, X = maximum and N = normal, but I haven't been able to create a `defS` on the Linux command line.

Anyway, all that was to say that I want to make sure the `gp_flags` do something sensible when the compression level is set, so I'll look at that too and submit an updated PR.

lib/zip/entry.rb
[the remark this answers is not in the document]
hainesr: Another thing: with the change I've already made, which would be reinforced by this change as well, the `put_next_entry` form which uses separate parameters instead of just an `Entry` is not used at all within the library itself. I do wonder if we could get rid of it altogether if we're making major version changes anyway at this point?

I'm going to think a bit more on that idea.

hainesr:
I've had a go at switching `Entry.new` to use named parameters. I'd appreciate comments, @jdleesmiller, as I've made a couple of calls that probably need discussion.

So I went for this,
```ruby
def initialize(zipfile = '', name = '', comment = '',
               compression_method = DEFLATED,
               compression_level = ::Zip.default_compression)
```
which pretty much keeps the order of the original parameters, but omits the more generally 'internal' ones, such as `crc`.

This means that when one does want to fully specify an `Entry`, which probably only really happens internally to the library, you have to do something like this:
```ruby
e = Entry.new(@name, entry.to_s, comment, compression_method, compression_level)
e.extra = extra
e.compressed_size = compressed_size
e.crc = crc
e.size = size
e.time = time
```

I _think_ this is probably fine as most of those 'internal' fields are only set when reading an existing entry from a Zip file.

Thoughts?

Other things that this PR now does is set the general purpose flags correctly for different compression levels, and it also sets the compression method to Store if compression level zero is specified. This last one is almost certainly the right thing to do, but may be considered a significant change.

Lastly, just about everything in this PR is a breaking change, so will require a bump to version 3 - should we set up a version 3 branch and collect other breaking changes there for now (if there are any)?

lib/zip/entry.rb
```diff
@@ -52,25 +53,28 @@ def check_name(name)
...
-      @comment            = args[2] || ''
-      @extra              = args[3] || ''
-      @compressed_size    = args[4] || 0
-      @crc                = args[5] || 0
-      @compression_method = args[6] || ::Zip::Entry::DEFLATED
-      @size               = args[7] || 0
-      @time               = args[8] || ::Zip::DOSTime.now
+      @zipfile            = zipfile
+      @comment            = comment
+      @compression_method = (@ftype == :directory ? STORED : compression_method)
```
jdleesmiller, line 68: I think the ternary here isn't needed, because `@compression_method` is only used if this is a file (and `@compression_level > 0`, which is missing from this check and the one in the setter but present in the getter). It might be clearer to call this something like `@requested_compression_method` to make it clear it's only used in some cases?

lib/zip/output_stream.rb
```diff
@@ -95,14 +95,14 @@ def put_next_entry(entry_name, comment = nil, extra = nil, compression_method =
       new_entry = if entry_name.kind_of?(Entry)
                     entry_name
                   else
-                    Entry.new(@file_name, entry_name.to_s)
+                    e = Entry.new(@file_name, entry_name.to_s, comment,
+                                  compression_method, level)
+                    e.extra = extra
```
jdleesmiller, line 100: Previously this assignment would only happen if `extra` was non-`nil`. It looks like everything still goes through OK with the new setter, but it relies `nil.kind_of?` and `nil.to_s`, which is fine but a bit tricky, so I thought I should check whether this was the intention.

test/central_directory_test.rb
```diff
@@ -57,10 +57,13 @@ def test_write_to_stream
...
     ::Zip.write_zip64_support = true
-    entries = [::Zip::Entry.new('file.zip', 'file1-little', 'comment1', '', 200, 101, ::Zip::Entry::STORED, 200),
-               ::Zip::Entry.new('file.zip', 'file2-big', 'comment2', '', 18_000_000_000, 102, ::Zip::Entry::DEFLATED, 20_000_000_000),
-               ::Zip::Entry.new('file.zip', 'file3-alsobig', 'comment3', '', 15_000_000_000, 103, ::Zip::Entry::DEFLATED, 21_000_000_000),
-               ::Zip::Entry.new('file.zip', 'file4-little', 'comment4', '', 100, 104, ::Zip::Entry::DEFLATED, 121)]
+    entries = [
+      ::Zip::Entry.new('file.zip', 'file-tiny', 'comment', Zip::Entry::STORED),
+      ::Zip::Entry.new('file.zip', 'file-big', 'comment'),
+      ::Zip::Entry.new('file.zip', 'file-alsobig', 'comment'),
+      ::Zip::Entry.new('file.zip', 'file-little', 'comment')
```
jdleesmiller, line 64: It looks like these test entries have lost their sizes. From the last line of the test, it seems that the local header offsets are enough to force zip64, but it seems like the test is perhaps a bit less realistic without the sizes?

test/entry_test.rb
```diff
@@ -40,42 +41,20 @@ def test_is_directory_and_is_file
...
-                              ::Zip::Entry::DEFLATED, 10_000)
+                              ::Zip::Entry::DEFLATED, ::Zip.default_compression)
     entry3 = ::Zip::Entry.new('file.zip', 'name2', 'isNotComparedXXX',
-                              'something extra', 123, 1234,
-                              ::Zip::Entry::DEFLATED, 10_000)
+                              ::Zip::Entry::DEFLATED, ::Zip.default_compression)
     entry4 = ::Zip::Entry.new('file.zip', 'name2', 'isNotComparedXXX',
-                              'something extraXX', 123, 1234,
-                              ::Zip::Entry::DEFLATED, 10_000)
-    entry5 = ::Zip::Entry.new('file.zip', 'name2', 'isNotComparedXXX',
```
jdleesmiller, line 54: If I understand the test correctly, `entry4 != entry5` is checking that the comparison is based on the compressed size. While setting up `entry5` would now require setting the compressed size post-construction, I think it's still a valid test to do? (And so on for entries 6-8.)

lib/zip/entry.rb
```diff
@@ -678,6 +698,24 @@ def data_descriptor_size
       (@gp_flags & 0x0008) > 0 ? 16 : 0
     end
 
+    # For DEFLATED compression *only*: set the general purpose flags 1 and 2 to
+    # indicate compression level. This seems to be mainly cosmetic but they are
+    # generally set by other tools - including in docx files.
```
jdleesmiller, line 703: The comments and link to APPNOTE in the PR description were helpful --- maybe worth adding them here?

hainesr:
Ah ha, I did wonder while I was doing the Entry API if I should do "keyword arguments", or if that might be what you meant. On reflection I agree with you and I think that would be better - not least because it would mean fewer changes elsewhere in the code. I'll change that about and see how it looks.

lib/zip/entry.rb
hainesr: Agreed that the ternary isn't needed here.

I'll leave the name as it is though, as it does have a meaning for an `Entry` that has been read from a Zip file. So much of `Entry` is confused by the fact that it is used both before and after writing to disk.

lib/zip/output_stream.rb
hainesr: Erk. Good catch, thanks. I've changed the signature of `OutputStream#put_next_entry` so that `extra` doesn't default to `nil`, **and** I've made the setter more robust.

test/central_directory_test.rb
hainesr: These revert back to how they were with the keyword argument API.

test/entry_test.rb
hainesr: Again, these just revert too.

lib/zip/entry.rb
hainesr: Done.

hainesr:
I've made the above changes and used keyword args for `Entry.new`. Keen for more comments if there are any, but if this looks OK then I'll add a documentation commit before merging.

> I don't have anything else in particular planned for a 3.x release, so I'd be fine with working on master, but if you have some other things you think it would make sense to push out at the same time, happy to set up an integration branch.

I do have some thoughts around extra fields that I'm toying with, but now I've started looking in more depth that's a long way down the track. So yes, might as well get this out and no need for integration.

hainesr:
I think I've covered the above notes @jdleesmiller - please could you have another look at this?

I've also now updated the `README` and `Changelog`, and along the way I've tidied up things like the author list as well (which I've taken the liberty of adding us to while I was at it).

PR #450, opened 2020-08-13
enable truffle ruby in Travis CI

gogainda:
@eregon enabling truffle ruby support for rubyzip

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ rvm:
   - 2.6
   - 2.7
   - ruby-head
+  - truffleruby-head
 matrix:
   fast_finish: true
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rvm:
   - 2.7
   - ruby-head
   - truffleruby-head
+  - truffleruby
 matrix:
   fast_finish: true
   include:
@@ -27,6 +28,8 @@ matrix:
     - rvm: ruby-head
     - rvm: rbx-4
     - rvm: jruby-head
+    - rvm: truffleruby-head
+    - rvm: truffleruby
 before_install:
   - gem --version
 before_script:


eregon:
```
  1) Error:
ZipFsFileNonmutatingTest#test_popen:
Errno::ENOMEM: Cannot allocate memory - posix_spawnp ls
    <internal:core> core/truffle/process_operations.rb:455:in `spawn'
    <internal:core> core/truffle/process_operations.rb:110:in `spawn'
    <internal:core> core/io.rb:894:in `popen'
    /home/travis/build/rubyzip/rubyzip/test/filesystem/file_nonmutating_test.rb:475:in `test_popen'
```

It looks like it's running out of memory while running tests.

hainesr:
Thanks for this. Please could you add truffleruby-head to the list of tests allowed to fail? Is there a stable truffleruby version that we would more reasonably expect to always work?

gogainda:
@eregon pls correct me if I am wrong, but we are not expecting truffleruby head to fail. It's should be as stable as "stable" version

eregon:
@gogainda Could you try if `truffleruby` (= latest release) has the same failure?
`truffleruby-head` is fairly stable, but still it changes regularly while `truffleruby` only changes once every 3 months.
So it's up to the gem maintainers to decide if they want to test on `truffleruby`, `truffleruby-head` or both.

In any case, could you file a bug on https://github.com/oracle/truffleruby/issues about this `Errno::ENOMEM` so we can track it on TruffleRuby's side?

hainesr:
I think we should test on both `truffleruby` and `truffleruby-head`, but let's have the head version allowed to fail in keeping with other rubies. Then the head can act as a "canary" for us without breaking the build if it does wobble. It should be allowed to wobble occasionally 😃

gogainda:
@eregon updated

eregon (commented):
Looks good to me.

jdleesmiller:
Thanks, works for me.

PR #452, opened 2020-08-24
Add support for Truffleruby in CI

gogainda:


gogainda:
@eregon

eregon:
Could you update https://github.com/rubyzip/rubyzip/pull/450 instead so we keep the conversation and diff together?
Also both `truffleruby` and `truffleruby-head` should be in `allow_failures` until https://github.com/oracle/truffleruby/issues/2076 is fixed.

PR #467, opened 2021-01-26
Fix depreciation warning: abstract_output_stream_test.rb:60: warning: `$,' is deprecated

bbuchalter:
Hello rubyzip community! 👋 

I was thinking about making a small contribution in another area, so I went to run the test suite and noticed this deprecation warning: `abstract_output_stream_test.rb:60: warning: '$,' is deprecated`.

I learned that setting `$,` (aka `$OUTPUT_FIELD_SEPERATOR`) to a non-nil value is now deprecated. This was discussed in https://bugs.ruby-lang.org/issues/14240 and implemented in ruby/ruby@6298ec2.

As a result, I've modified both some test code and application code to assume a world in which this value is always `nil`. 

Since I'm new to this community, and this code [was originally almost 20 years ago(!)](https://github.com/rubyzip/rubyzip/commit/441dcb964393fcb43a4a507f8ff6402cf93ed8fb#diff-859f6edc8864e2e7b8ed6fad65231c59c3f648cc645f1d2ee81f73ca58f49d6fR401), perhaps we should leave it as is until forced to do otherwise?

Anyway, just a little contribution that I hope is found helpful. Let me know if a CHANGELOG entry would be expected for a change this small.

### DIFF
--- a/test/ioextras/abstract_output_stream_test.rb
+++ b/test/ioextras/abstract_output_stream_test.rb
@@ -20,12 +20,10 @@ class AbstractOutputStreamTest < MiniTest::Test
   def setup
     @output_stream = TestOutputStream.new
 
-    @save_comma_sep = $OUTPUT_FIELD_SEPARATOR
     @save_output_sep = $OUTPUT_RECORD_SEPARATOR
   end
 
   def teardown
-    $OUTPUT_FIELD_SEPARATOR = @save_comma_sep
     $OUTPUT_RECORD_SEPARATOR = @save_output_sep
   end
 
@@ -57,7 +55,6 @@ class AbstractOutputStreamTest < MiniTest::Test
     @output_stream.print('I sure hope so!')
     assert_equal("hello world. You ok out there?\nI sure hope so!\n", @output_stream.buffer)
 
-    $OUTPUT_FIELD_SEPARATOR = nil
     @output_stream.buffer = ''
     @output_stream.print('monkey', 'duck', 'zebra')
     assert_equal("monkeyduckzebra\n", @output_stream.buffer)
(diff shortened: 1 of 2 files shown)


test/ioextras/abstract_output_stream_test.rb
```diff
@@ -50,19 +48,18 @@ def test_print
...
     assert_equal("hello world. You ok out there?\n", @output_stream.buffer)
 
     @output_stream.print('I sure hope so!')
     assert_equal("hello world. You ok out there?\nI sure hope so!\n", @output_stream.buffer)
 
-    $, = 'X'
     @output_stream.buffer = ''
     @output_stream.print('monkey', 'duck', 'zebra')
-    assert_equal("monkeyXduckXzebra\n", @output_stream.buffer)
+    assert_equal("monkeyduckzebra\n", @output_stream.buffer)
```
hainesr, line 60: I guess this test doesn't make any sense now? It's there purely to test that `AbstractOutputStream` respects `$,`. We have a test above (`@output_stream.print(' You ok ', 'out ', 'there?')`) which tests `#print` with multiple strings.

Should we remove this?

hainesr (changes requested):
Thanks for this. Sorry it's taken so long for me to look at it.

This looks obviously sensible I think and now would be a good time to implement with a major version bump on the way. Just one question about a now possibly unneeded test...

test/ioextras/abstract_output_stream_test.rb
bbuchalter: I think this PR can ship without this further change. My intention was simply to remove the deprivation which this PR accomplishes. If we want to remove this test, we can do so in another PR. 

Personally, I don't have an opinion on whether or not the test should be removed.

hainesr: OK. I had to resolve a conflict on those lines anyway, so I got rid of them then.

hainesr:
Thanks for the PR @bbuchalter - all merged in now.

PR #474, opened 2021-03-05
add Ruby 3.0 to CI

taichi-ishitani:
This PR is to add Ruby 3.0 to CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ rvm:
   - 2.5
   - 2.6
   - 2.7
+  - 3.0
   - ruby-head
   - truffleruby-head
   - truffleruby


clarkedb:
I see this adds Ruby 3.0 to CI, but will `rubyzip` 3.0.0 be the first release that supports ruby 3.0? Or does 2.3.2 already support it?

hainesr:
That's a good question. Our CI on the `2.x` branch died with Travis. I'll see about getting GitHub Actions backported and see where we are with that...

Thanks for raising this.

petergoldstein:
@hainesr At this point we probably also want to add Ruby 3.1.

hainesr:
Actions backported and confirmed that rubyzip 2.3.2 works with ruby 3.0 and 3.1. Note added to the README.

PR #475, opened 2021-03-06
fix frozen string literal error

taichi-ishitani:
I got `frozen string literal error` during running `rake test` with the `--enable-frozen-string-literal` option.
This PR is to fix these errors.


aliismayilov:
Funny that I was also going to submit similar PR :)

Does it make sense to:
1. Add `frozen_string_literal: true` magic comment on top of the ruby source files, so that even the projects that doesn't use `--enable-frozen-string-literal` would benefit from the memory usage reduction?
2. Add `RUBYOPT="--enable-frozen-string-literal"` in the travic ci file, so that in the future developers don't modify strings accidentally as the test suite would catch it.

hainesr:
Many thanks for this. I have merged this commit directly into a branch I am building to update rubocop and fix some of the offences. As part of this work I have added the magic comment to the top of all the files. Your commit will be merged into master as part of this soon I hope.

Do we need to set `RUBYOPT="--enable-frozen-string-literal"` in the tests if the magic comment is in place and rubocop is checking for the presence of it? The CI fails if rubocop fails.

Thanks again!

hainesr:
Merged, thanks!

PR #476, opened 2021-03-09
Fix builds

aliismayilov:
master branch builds seem to be broken. This PR updates the tests that should have been updated in https://github.com/rubyzip/rubyzip/commit/a0345420d87d781e7bcef0b78f3d7ed785fedcfb


hainesr:
The tests have already been fixed on a different PR (#473) that then doesn't rely on fixed sizes, and so should be more robust.

Issue #482, opened 2021-05-25
Refactor `assert_forwarded` in the tests?


hainesr:
It's a horrible bit of code which obfuscates what it is doing; it relies on an API which is not recommended (`_id2ref`); it requires that `ObjectSpace` be enabled to run the JRuby tests; and, it currently breaks builds in TruffleRuby (albeit due to a bug in TruffleRuby oracle/truffleruby#2358).

Can we be rid of it?

eregon:
I'll take a shot at this.

eregon:
https://github.com/rubyzip/rubyzip/pull/483

PR #483, opened 2021-05-25
Refactor assert_forwarded so it does not need ObjectSpace._id2ref or eval

eregon:
Fixes https://github.com/rubyzip/rubyzip/issues/482


hainesr:
Wow, many thanks for this @eregon!

PR #486, opened 2021-05-31
Fix restore options consistency.

### LINKED ISSUE #336: View and/or preserve original date created, date modified?
For files inside the zip file, is there any way in rubyzip to access their original dates created and modified? If I extract them, Windows shows them as having been created and modified at the time of the extraction, which is quite different from the dates showing inside the zip file. I need to be able to list the dates they were originally created and last modified and, ideally, preserve those dates with the extracted files.

I have searched for this answer both here and on the Web, and the only references to solutions I can find use DotNetZip or Python. I would like to use Ruby.

Thanks.

### LINKED ISSUE #422: Inconsistent behaviour of `Zip::File#get_entry` and `Zip::File#find_entry` exposes options out of sync.

hainesr:
This PR is, finally, to finish off the work to fix and make consistent the options to restore file timestamps and file permissions on extraction.

The default options are now specified at the top level so that both `File` and `Entry` get a consistent view of them. They can still be overridden per-`File` - and per-`Entry` if needs be.

Additionally this PR enables timestamps to be restored on Windows.

This PR closes #422 and fixes #336.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 # 3.0.0 (Next)
 
+- Fix restore options consistency. [#486](https://github.com/rubyzip/rubyzip/pull/486)
+- View and/or preserve original date created, date modified? (Windows). [#336](https://github.com/rubyzip/rubyzip/issues/336)
 - Fix frozen string literal error. [#475](https://github.com/rubyzip/rubyzip/pull/475)
 - Set the default `Entry` time to the file's mtime on Windows. [#465](https://github.com/rubyzip/rubyzip/issues/465)
 - Ensure that `Entry#time=` sets times as `DOSTime` objects. [#481](https://github.com/rubyzip/rubyzip/issues/481)
--- a/lib/zip.rb
+++ b/lib/zip.rb
@@ -48,6 +48,12 @@ module Zip
                 :force_entry_names_encoding,
                 :validate_entry_sizes
 
+  DEFAULT_RESTORE_OPTIONS = {
+    restore_ownership:   false,
+    restore_permissions: false,
+    restore_times:       false
+  }.freeze
+
   def reset!
     @_ran_once = false
     @unicode_names = false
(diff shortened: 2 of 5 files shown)


hainesr:
I have submitted this as a PR, rather than just merging it, as it is a change in defaults, etc.

jspanjers:
@hainesr , could you pick the last to commits from https://github.com/rubyzip/rubyzip/pull/490 to fix the Windows tests?

PR #489, opened 2021-06-06
Fix windows tests (revised)

### LINKED ISSUE #485: Many tests failing under Windows.
A lot of failures appear to be due to hardcoded paths into places like `/tmp` which should be easy to fix. Other tests appear to fail due to unexpected line endings - `\r\n` vs `\n` - which I think suggests that we're testing what comes out of a zip file (`\n`) with a text file that has been auto-converted to `\r\n` in the checkout process?

jspanjers:
Fixes #485

### DIFF
--- a/lib/zip/entry.rb
+++ b/lib/zip/entry.rb
@@ -164,8 +164,9 @@ module Zip
       return false unless cleanpath.relative?
 
       root = ::File::SEPARATOR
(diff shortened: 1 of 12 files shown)


hainesr (approved):
Thank you so much for this @jspanjers!

PR #490, opened 2021-06-06
Fix restore options consistency (rebased)

jspanjers:
Rebased pull request 486 on current master, fixing Windows tests.


hainesr:
Thanks for this. I have cherrypicked your two commits from here into the original PR as I've made other changes to master as well.

I'll close this now - thanks for the commits!

PR #495, opened 2021-06-26
Remove duplicated line on Changelog.md

nishidayuya:
Line 15 is same as line 18.

Line 15 (at 25795c7b 2021-05-18) is newer than line 18 (at 0a6037b0 2020-11-08), so this pull-request removes line 15.


hainesr:
Good spot, thanks!

PR #498, opened 2021-08-03
Updating readme for password protection

dubsdaazn:
Current initialization of `OutputStream` does not use keyword parameters

https://github.com/rubyzip/rubyzip/blob/bc6523ec43e911d7931aafbda2a81dae68617a90/lib/zip/output_stream.rb#L29


dubsdaazn:
Just realize that master uses keyword parameters but the latest version 2.3.2 isn't current with master

PR #504, opened 2021-11-07
Fix `OutputStream#put_next_entry` to preserve `StreamableStream`s.

### LINKED ISSUE #503: Can't set time with get_output_stream
I'm trying to put an entry with some arbitrary timestamp into the zip file, but it doesn't work, it always writes `Time.now`:
```ruby
require 'zip'
time = Time.new '2020-01-01' # some random date
Zip::File.open 'foo.zip', create: true do |zip|
  e = Zip::Entry.new zip.name, 'some_file'
  e.time = Zip::DOSTime.from_time time
  zip.get_output_stream(e) {|s| s.puts 'foobar' }
end
```
But unzip reports the current date:
```
$ unzip -l foo.zip 
Archive:  foo.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        7  11-06-2021 17:56   some_file
---------                     -------
        7                     1 file
```

It looks like `File#get_output_stream` adds a `StreamableStream` to `@entry_set`, but later in `OutputStream#put_next_entry` it checks `entry_name.kind_of?(Entry)`, which will be false and it creates a new `Entry` ignoring everything I've set up in the entry. The same happens if I don't create the entry manually but pass the time to `#get_output_stream`.
`File#add` works correctly, but it requires creating a temporary file and manually setting the last modification date on it.
Using ruby 3.0.2p107 and rubyzip 2.3.2.

hainesr:
When passing an `Entry` type to `File#get_output_stream` the entry is
used to create a `StreamableStream`, which preserves all the info in the
entry, such as timestamp, etc. But then in `put_next_entry` all that is
lost due to the test for `kind_of?(Entry)` which a `StreamableStream` is
not. See #503 for details.

This change tests for `StreamableStream`s in `put_next_entry` and uses
them directly. Some set-up within `Entry` needed to be made more robust
to cope with this, but otherwise it's a low impact change, which does
fix the problem.

The reason this case was being missed before is that the tests weren't
testing `get_output_stream` with an `Entry` object, so I have also added
that test too.

Fixes #503.

### DIFF
--- a/lib/zip/entry.rb
+++ b/lib/zip/entry.rb
@@ -83,14 +83,21 @@ module Zip
       @ftype = name_is_directory? ? :directory : :file
 
       @zipfile            = zipfile
-      @comment            = comment
-      @compression_method = compression_method
-      @compression_level  = compression_level
+      @comment            = comment || ''
+      @compression_method = compression_method || DEFLATED
+      @compression_level  = compression_level || ::Zip.default_compression
 
-      @compressed_size    = compressed_size
-      @crc                = crc
-      @size               = size
-      @time               = time
+      @compressed_size    = compressed_size || 0
+      @crc                = crc || 0
+      @size               = size || 0
+      @time               = case time
+                            when ::Zip::DOSTime
+                              time
+                            when Time
+                              ::Zip::DOSTime.from_time(time)
+                            else
+                              ::Zip::DOSTime.now
+                            end
       @extra              =
         extra.kind_of?(ExtraField) ? extra : ExtraField.new(extra.to_s)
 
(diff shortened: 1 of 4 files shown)


u3shit:
>The reason this case was being missed before is that the tests weren't
>testing get_output_stream with an Entry object, so I have also added
>that test too.

I don't think that is the case, since `get_output_stream` creates an `Entry` internally then wraps it in a `StreamableStream` even when you don't pass an `Entry` to it. As far as I can see the in memory `File` is okay, only writing to a file is broken (before this PR). `test_get_output_stream` calls `get_entry` which just gets the entry from `@entry_set`, while `test_get_output_stream_with_entry` actually saves the zip to disk and reloads it, so it also checks `OutputStream`.

PR #517, opened 2022-02-01
Enable Zip64 by default

stanhu:
Previously if RubyZip attempted to create an archive with more than
64K entries, the central directory would truncate the count. `unzip`
and `zipinfo` would fail with an error message such as:

```
error:  expected central file header signature not found (file #93272).
  (please check that you have transferred or created the zipfile in the
  appropriate BINARY mode and that you have compiled UnZip properly)
```

This generated a lot of confusion and a production issue since many
tools fail to decode a RubyZip-created archive if Zip64 is not enabled
for a large number of files. Since Zip64 support is now the norm,
enable this by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,10 +328,10 @@ end
 
 ### Zip64 Support
 
-By default, Zip64 support is disabled for writing. To enable it do this:
+By default, Zip64 support is enabled for writing. To disable it do this:
 
 ```ruby
-Zip.write_zip64_support = true
+Zip.write_zip64_support = false
 ```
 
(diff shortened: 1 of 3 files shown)


hainesr:
Thanks for this @stanhu.

I am going to park this for a bit while I check that setting `Zip.write_zip64_support = true` by default doesn't have unintended side-effects to the files produced. I know the tests pass, but there is some mucking about with the extra fields that goes on and it's possible that we need to update how we do Zip64 to avoid this. I'm concerned that by turning this on by default we leave placeholders in the files that we don't need if a file doesn't actually need the Zip64 support.

Anyway, I will look into this, and I do agree that we should support Zip64 by default when it is required, rather than needing to turn it on specifically.

PR #522, opened 2022-04-06
Fix indentation in sample code

ham0215:
The indentation of the sample code in the README was incorrect, so I fixed it.


hainesr:
Good spot, thanks!

Issue #523, opened 2022-04-21
Zip::DecompressionError:        zlib error while inflating


308820773:
```ruby

string_buffer = Zip::OutputStream.write_buffer(::StringIO.new('test'), Zip::TraditionalEncrypter.new('1')) do |out|
  out.put_next_entry('1.test')
  out.write string_io.read
end.string

zip_file = Tempfile.new(["1", ".zip"])
zip_file.binmode
zip_file.puts string_buffer


Zip::InputStream.open(::StringIO.new(zip_file.open.read), 0, Zip::TraditionalDecrypter.new("1")) do |input|
  entry = input.get_next_entry
  name  = entry.name.force_encoding('utf-8')
  File.open(name, 'w+') { |f| f.puts entry.get_input_stream.read }
end
```

hainesr:
Which version of Ruby and Rubyzip are you using?

Thanks.

308820773:
ruby 2.6.6p146
rubyzip (2.3.2)

308820773:
https://github.com/rubyzip/rubyzip/pull/524

I check the code, get_input_stream miss the Decrypter. When add the Decrypter it's ok!

308820773:
Expect 3.0

hainesr:
Thanks. The encryption/decryption in rubyzip is patchy at the moment. I'll see what I can do quickly for the 3.0 release.

hainesr:
OK, I've looked at this properly now. It's not a bug in Rubyzip, but we are missing some documentation.

When using `Zip::InputStream#get_next_entry` the behaviour is to return the next entry in the stream and _position the stream at the start of the entry data_. So your reading code should look like the following instead:

```ruby
Zip::InputStream.open(::StringIO.new(zip_file.open.read), 0, Zip::TraditionalDecrypter.new("1")) do |input|
  entry = input.get_next_entry
  name  = entry.name.force_encoding('utf-8')
  File.open(name, 'w+') { |f| f.puts input.read }
end
```

I'll update the ReadMe to make this obvious.

hainesr:
Docs updated.

PR #524, opened 2022-04-22
fix get_input_stream no decrypter

308820773:


hainesr:
This isn't required as the APIs already exist to read encrypted zip archives. See my comment here: https://github.com/rubyzip/rubyzip/issues/523#issuecomment-1107448443

Issue #530, opened 2022-06-13
`Zip::FileSystem::ZipFsFile#symlink?` always returns false


fbacall:
even if the entry for the given filepath is actually a symlink

https://github.com/rubyzip/rubyzip/blob/master/lib/zip/filesystem/file.rb#L200-L202

fbacall:
I wonder if the behaviour of `ftype` should also be changed?

https://github.com/rubyzip/rubyzip/blob/master/lib/zip/filesystem/file.rb#L208-L210

The `ftype` property of a `Zip::Entry` https://github.com/rubyzip/rubyzip/blob/044759f50263f9263461fffc65ab2c364301c9ad/lib/zip/entry.rb#L379-L407 can return `:symlink` (as a symbol)

But on the other hand, I wonder if this method is supposed to behave like Ruby's `ftype` on `File.stat` which returns a `link` (as a string): https://ruby-doc.org/core-2.5.0/File/Stat.html#method-i-ftype

:thinking:

hainesr:
That would seem sensible. Did `File.stat` change at any point, do you know?

I suppose if we were to do this then now would be the time, with version 3.0 coming up (at some point 😬 ).

hainesr:
Having now looked this up, `ftype` has been returning strings since at least 1.9.3...

hainesr:
I agree: let's not open the symlinks can of worms right now - although I'll have to get to it at some point 😢 

Do you think we should change `ftype` to return strings though, to match `File.stat`? I'm leaning towards yes. I also note that there are side effects in `set_ftype_from_c_dir_entry` that I don't much care for (c.f. setting `@unix_perms`)...

fbacall:
It makes sense to me, but I'm not sure how much of a breaking change that would be. As well as changing to strings, it would also need to return `link` instead of `symlink` for symlinks.

hainesr:
It would be a breaking change, but now is the time for those because next release is v3.0.

I'll see about doing this soon.

PR #531, opened 2022-06-14
Implement `Zip::FileSystem::ZipFsFile#symlink?`

fbacall:
Addresses #530

### DIFF
--- a/test/filesystem/file_nonmutating_test.rb
+++ b/test/filesystem/file_nonmutating_test.rb
@@ -189,6 +189,7 @@ class FileNonmutatingTest < MiniTest::Test
     zip_file = ::Zip::File.new('test/data/path_traversal/tuzovakaoff/symlink.zip')
     assert(zip_file.file.symlink?('path'))
     assert(!zip_file.file.symlink?('path/file.txt'))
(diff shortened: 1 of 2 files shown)


test/filesystem/file_nonmutating_test.rb
```diff
@@ -186,7 +186,10 @@ def test_blockdev?
   end
 
   def test_symlink?
-    assert_always_false(:symlink?)
+    zip_file = ::Zip::File.new('test/data/path_traversal/tuzovakaoff/symlink.zip')
+    assert(zip_file.file.symlink?('path'))
+    assert(!zip_file.file.symlink?('path/file.txt'))
```
hainesr, line 191: Please can you use refute here instead of negating the condition?

hainesr (changes requested):
Hi Finn,

Thanks for this - just one comment!

Rob

test/filesystem/file_nonmutating_test.rb
hainesr: Ah, ok. I haven't cleaned that one up yet 😬 

Leave it then and I'll sort the whole file out at some point. 

Cheers

PR #534, opened 2022-08-04
Fix unraised error on encrypted archives

Brcrwilliams:
Currently, we raise an error for encrypted zip entries if `@decrypter` is a `NullEncryper`. However, the default value for `@decrypter` is actually `NullDecrypter`.


hainesr:
Nice catch, thanks! 🚀

PR #537, opened 2022-09-16
Adding an extract zip example

Atout-Agile:
Adding a secttion "extract zip" in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,36 @@ Zip::File.open('foo.zip') do |zip_file|
 end
 ```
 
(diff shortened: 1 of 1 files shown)


hainesr:
Hello. Thanks for this and apologies for taking so long to look at your PR.

What does this addition do that the code example in the README just above it doesn't do? I guess this example is more explicit about the destination path of the extracted entry. Would it not be better to add this, and/or perhaps an explanation, to the code that is already there, rather than add a whole new example?

PR #541, opened 2022-11-18
Make Zip64 default for writing, but only where it is needed

### LINKED ISSUE #521: Consider enabling write_zip64_support by default?
I accidentally created an invalid zip file by adding a single file with > 4GB of uncompressed data but without setting `Zip.write_zip64_support = true`. For example, `7z x` errors when attempting to extract the generated file.

In my case I quickly found the problem and recreated the zip file having enabled `write_zip64_support`, but I can imagine in some scenarios it may be a pain to (much?) later find that the generated file has been silently corrupted.

Is there a good reason _not_ to enable `write_zip64_support` by default? Or perhaps explicitly raise an error if `rubyzip` detects that it has written >4GB of data?

A simple reproduction:
```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rubyzip', "= 2.3.2"
end

require "zip"

ONE_MB_OF_A = "a" * 1024 * 1024

def write_zip(file_name, zip64_support:)
  Zip.write_zip64_support = zip64_support

  Zip::File.open(file_name, Zip::File::CREATE) do |zipfile|
    zipfile.get_output_stream("data.txt") do |data|
      (1024 * 4).times do |i|
        data << ONE_MB_OF_A
      end

      10.times do |i|
        data << "b"
      end
    end
  end
end

write_zip("example.zip", zip64_support: false)
write_zip("example64.zip", zip64_support: true)
```
writes out 4GB + 10 bytes of data, running `7z t example.zip` fails but `7z t example64.zip` passes

hainesr:
This PR builds on #517.

It turns on Zip64 write support by default (thanks @stanhu, I've merged your PR into this one), and adds features to ensure that the Zip64 extra data is only added when it is needed.

Fixes #521.

Please can anyone interested in this feature (@stanhu, @owst) test and check it does what you expect? I intend to merge this in for version 3 (which is getting closer, honest).


owst:
Hi @hainesr, thanks! It seems to work well with a trivial test:

```ruby
require "zip"

ONE_MB_OF_A = "a" * 1024 * 1024

def write_zip(file_name)
  Zip::File.open(file_name, create: true) do |zipfile|
    zipfile.get_output_stream("data.txt") do |data|
      (1024 * 4).times do |i|
        data << ONE_MB_OF_A
      end

      10.times do |i|
        data << "b"
      end
    end
  end
end

write_zip("example.zip")
```

Checking with `7zz t example.zip` confirms the file is well-formed:
```
Testing archive: example.zip
--
Path = example.zip
Type = zip
Physical Size = 4817063

Everything is Ok

Size:       4294967306
Compressed: 4817063
```

So that looks good - thank you for your work on this PR!

hainesr:
Thanks for testing @owst 🚀

PR #545, opened 2023-01-16
Zip::InputStream#read(0) should return an empty string

sakuro:
`Zip::InputStream#read` returns `nil` when `number_of_bytes` is `0`.

It is not compatible with [`IO#read(0)`](https://ruby-doc.org/3.2.0/IO.html#method-i-read:~:text=If%20maxlen%20is%20zero%2C%20returns%20an%20empty%20string.) which returns an empty string.

```rb
open("/etc/passwd") {|io| io.read(0) }
#=> ""
```

### DIFF
--- a/lib/zip/ioextras/abstract_input_stream.rb
+++ b/lib/zip/ioextras/abstract_input_stream.rb
@@ -36,7 +36,7 @@ module Zip
                end
 
         if tbuf.nil? || tbuf.empty?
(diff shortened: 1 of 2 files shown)


hainesr:
Thank you for this, and apologies for how long it's taken me to look at it properly.

I have merged this into my v3 prep branch, so it will hit master soon, hopefully.

PR #549, opened 2023-01-30
Fix typo know -> known

pboling:
Single char fix in readme.


hainesr:
Thanks for this. I have merged it into a branch where we are preparing version 3, which will be merged into master when ready.

PR #550, opened 2023-03-06
add non-deterministic breaking test

machty:
See https://github.com/titusfortner/webdrivers/issues/245

I believe I've found a non-deterministic bug related to YJIT + rubyzip. This PR has a clumsy breaking test that should trigger the issue I'm seeing in my own projects. I know this isn't "clean" yet but please approve the Github Action workflow to at least verify the issue.

The breaking test does the following:

1. Downloads `chromedriver` zip
2. Repeatedly attempts to run a ruby script that unzips the file using Rubyzip, with YJIT enabled
3. Expected behavior: it will segfault.

I'm finding that the reason it segfaults is that the binary file produced by the decompression is not reliably the same size (it's off by a few bytes).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -63,6 +63,27 @@ jobs:
         ruby-version: ${{ matrix.ruby }}
         bundler-cache: true
 
+    - name: Chromedriver non-deterministic test
+      run: |
+        curl "https://chromedriver.storage.googleapis.com/110.0.5481.77/chromedriver_linux64.zip" > chromedriver_linux64.zip
+        rm -f chromedriver && RUBY_YJIT_ENABLE=true bundle exec ruby test/yjit_test.rb && chmod +x ./chromedriver && ./chromedriver --version
+        rm -f chromedriver && RUBY_YJIT_ENABLE=true bundle exec ruby test/yjit_test.rb && chmod +x ./chromedriver && ./chromedriver --version
+        rm -f chromedriver && RUBY_YJIT_ENABLE=true bundle exec ruby test/yjit_test.rb && chmod +x ./chromedriver && ./chromedriver --version
+        rm -f chromedriver && RUBY_YJIT_ENABLE=true bundle exec ruby test/yjit_test.rb && chmod +x ./chromedriver && ./chromedriver --version
+        rm -f chromedriver && RUBY_YJIT_ENABLE=true bundle exec ruby test/yjit_test.rb && chmod +x ./chromedriver && ./chromedriver --version
+        rm -f chromedriver && RUBY_YJIT_ENABLE=true bundle exec ruby test/yjit_test.rb && chmod +x ./chromedriver && ./chromedriver --version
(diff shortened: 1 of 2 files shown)


machty:
Hmm, unfortunately I have not been able to reproduce this failure within the rubyzip repo :(

Feel free to close this PR -- I'm leaving it open for now in case it resonates with anyone.

alexplatteeuw:
Not sure it's the same problem but we also run into errors recently. I was able to reproduce on ubuntu with ruby 3.2 YJIT (it's working fine on macos or working fine without YJIT).

The errors we witnessed are not always the same. Sometimes we have:

image

Sometimes the file is unzipped but the content is damaged (the beginning of the file is fine but the end is a mess with invalid UTF-8 bytes).

I would be happy to give more details. It seems someone is facing the same issue [here](https://github.com/Shopify/yjit/issues/310#issuecomment-1457566139)

hainesr:
Thank you for this, and many apologies that it took me so long to be able to look at it.

It looks like you've got to the bottom of this between you, but I've run the tests anyway and it looks like it's as expected.

Shall I close this now?

Thanks again all!

maximecb:
The bug in YJIT is fixed.

machty:
@maximecb Is the bugfix in Ruby 3.2.2? I tried upgrading to 3.2.2 and was able to reproduce the error in my Github Actions CI

```
Using 3.2.2 as input from file .ruby-version
...

2023-04-10 11:51:34 DEBUG Webdrivers Get response: #<Net::HTTPOK 200 OK readbody=true>
2023-04-10 11:51:34 DEBUG Webdrivers Successfully downloaded /tmp/20230410-2755-z3r9atchromedriver_linux64.zip
2023-04-10 11:51:34 DEBUG Webdrivers Decompressing /tmp/20230410-2755-z3r9atchromedriver_linux64.zip
rails aborted!
Zip::EntrySizeError: entry 'chromedriver' should be 13[97](https://github.com/FutureProofRetail/business-server/actions/runs/4629952279/jobs/8241461500#step:18:98)7040B, but is larger when inflated.
/home/runner/work/business-server/business-server/vendor/bundle/ruby/3.2.0/gems/rubyzip-2.3.2/lib/zip/entry.rb:631:in `block (2 levels) in create_file'
/home/runner/work/business-server/business-server/vendor/bundle/ruby/3.2.0/gems/rubyzip-2.3.2/lib/zip/entry.rb:540:in `get_input_stream'
/home/runner/work/business-server/business-server/vendor/bundle/ruby/3.2.0/gems/rubyzip-2.3.2/lib/zip/entry.rb:621:in `block in create_file'
/home/runner/work/business-server/business-server/vendor/bundle/ruby/3.2.0/gems/rubyzip-2.3.2/lib/zip/entry.rb:620:in `open'
/home/runner/work/business-server/business-server/vendor/bundle/ruby/3.2.0/gems/rubyzip-2.3.2/lib/zip/entry.rb:620:in `create_file'
```

XrXr:
Unfortunately, the fix was not included in the 3.2.2 release. Here's the ticket: https://bugs.ruby-lang.org/issues/19483
The status for 3.2 does not say "DONE". @machty

PR #560, opened 2023-07-28
Add two Windows head builds to CI, update README.md CI info

MSP-Greg:
The setup-ruby action installs an MSYS2 version of zip, so the step running `choco install zip` isn't needed.


hainesr:
Many thanks for this. I don't have a Windows machine to test on, so this is most appreciated!

Rob

PR #568, opened 2024-02-16
Docs: Update init for decryptor and encryptor example

KamilDzierbicki:
A small update to the example of how to use the decrypter and the encrypter in this gem for v2.3.2.


hainesr:
Thanks @KamilDzierbicki.

I'm not going to merge this, sorry, because I don't want to lose the newer version information. So I have combined your update into https://github.com/rubyzip/rubyzip/commit/fd0cf5443ea95c781750f4c0ec735e6cec92ce38, thanks.

PR #572, opened 2024-03-20
Ensure compatibility with `--enable-frozen-string-literal`

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/20205

In Ruby 3.4 string literals will be "chilled" by default. Meaning they are still mutable, but will pretend to be frozen.

In most case it has no impact, just emit a few warnings there and there, but there is one thing it impacts is the `StringIO.new('')` pattern. `StringIO` checks if the given string is frozen, and if it is will act as a read only IO.

This breaks rubyzip 2.x.

This commit make the 2.x branch compatible with frozen string literals.

Note: I see all these issues are solved on master, but only released in an alpha, and I suppose ultimately will require a major version upgrade.

I think it would be good to provide a 2.x version compatible with Ruby 3.4, but your call.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu]
-        ruby: ['2.4', '2.5', '2.6', '2.7', '3.0', '3.1', jruby, truffleruby]
+        ruby: ['2.4', '2.5', '2.6', '2.7', '3.0', '3.1', '3.2', '3.3', jruby, truffleruby]
         include:
           - os: macos
             ruby: '2.4'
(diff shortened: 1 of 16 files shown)


hainesr:
Thanks for doing this.

I have to say my initial thought though was that I'd like to leave the 2.x branch (and Ruby 2.x) behind and frozen string literals would be a good way of doing this. However, it occurs to me that 3.x declares `frozen_string_literal: true` in every file, so it's certainly worth having 2.4 support this so that people can test before moving to 3.0.

I'll merge this to the 2.4 branch, thanks.

casperisfine:
Thank you very much 🙇

hainesr:
Merged into the 2.4 branch now.

Thanks again for your help!

hainesr:
Released in 2.4.rc1 on rubygems now!

miry:
It works, thank you very much.

PR #575, opened 2024-04-09
Fix `File#write_buffer` to always return the given `io`

casperisfine:
Ref: https://github.com/rubyzip/rubyzip/commit/ef89a62b70d0b0c43a9579d414d5a917fef0cdc3

This fixes a regression in 2.4.rc1.

cc @hainesr

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -412,7 +412,7 @@ module Zip
 
     # Write buffer write changes to buffer and return
     def write_buffer(io = ::StringIO.new)
-      return unless commit_required?
+      return io unless commit_required?
 
       ::Zip::OutputStream.write_buffer(io) do |zos|
         @entry_set.each { |e| e.write_to_zip_output_stream(zos) }
--- a/test/file_test.rb
+++ b/test/file_test.rb
@@ -141,7 +141,7 @@ class ZipFileTest < MiniTest::Test
     assert zf.entries.map(&:name).include?('zippedruby1.rb')
 
     # Ensure the buffer isn't changed.
-    zf.write_buffer(string_io)
+    assert_same string_io, zf.write_buffer(string_io)
     assert_equal(data, string_io.string)
   end
 


casperisfine:
Thanks for the quick merge. I have a few more failures in my app, I'll try to figure if they are regressions or us using undocumented behavior.

hainesr:
If you can stand it, it's probably worth seeing how things go with 3.0.0 too - there are like 3 years of bug fixes in there 😮 

I'm going to try and cut an RC for version 3 later today...

casperisfine:
>  it's probably worth seeing how things go with 3.0.0 too

yeah, I ought to look at it soonish. I attempted an upgrade and saw hundreds of errors on CI, so I've put it on the backlog 😅

casperisfine:
Alright, I got a minmal repro for what I think is a regression:

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rubyzip', ENV.fetch("VERSION", "2.3.2")
  gem 'minitest'
end


require 'zip'
require 'zip/version'

puts "RubyZip: #{Zip::VERSION}"

require 'minitest/autorun'

class ZipTest < Minitest::Test
  ZIP = "PK\x03\x04\x14\x00\x00\x00\b\x002^\x89X\x16\xE8v@c\x00\x00\x00\x01\x00\x01\x00\x14\x00\x00\x00resources_export.csv\xED\xC1\xC9\r\x00\x10\x00\x00\xB0\xBFYL%D<x8\x12\xE3\eD\xDB\x96\xE38=\x95\x19\xD7\x9Em\xD4p\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xF8QxPK\x01\x024\x03\x14\x00\x00\x00\b\x002^\x89X\x16\xE8v@c\x00\x00\x00\x01\x00\x01\x00\x14\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\xA4\x81\x00\x00\x00\x00resources_export.csvPK\x05\x06\x00\x00\x00\x00\x01\x00\x01\x00B\x00\x00\x00\x95\x00\x00\x00\x00\x00"

  def test_write_into_other_io
    other_io = StringIO.new

    Zip::File.open_buffer(ZIP) do |z|
      z.write_buffer(other_io)
    end

    other_io.rewind
    assert_equal ZIP.bytesize, other_io.size
  end
end
```

Basically, I think `write_to` should always write, because it can't assume it's being asked to write back into the same IO it read from.

Now maybe I misunderstand the API?

I'll open a PR and we can discuss there.

Issue #512, opened 2021-11-29
open_buffer corrupts the buffer it's given


Taywee:
This is maybe related to #280, but in my case, the archive doesn't just have its order changed, but it's in fact completely corrupted.  Here is a complete example, including the archive embedded (Just a simple zip with a file `foo.txt` with the contents `"FOO!\n"` and `bar.txt` with the contents `"BAR?\n"`):

```ruby
#!/usr/bin/env ruby

require 'zip'

archive = "PK\x03\x04\n\x00\x02\x00\x00\x00\x8Cr}S\x9F$j\xC5\x05\x00\x00\x00\x05\x00\x00\x00\a\x00\x1C\x00foo.txtUT\t\x00\x03\x17D\xA5a'D\xA5aux\v\x00\x01\x04\xE8\x03\x00\x00\x04\xE8\x03\x00\x00FOO!\nPK\x03\x04\n\x00\x02\x00\x00\x00\x90r}S\x90\xFA\x8A\x10\x05\x00\x00\x00\x05\x00\x00\x00\a\x00\x1C\x00bar.txtUT\t\x00\x03 D\xA5a'D\xA5aux\v\x00\x01\x04\xE8\x03\x00\x00\x04\xE8\x03\x00\x00BAR?\nPK\x01\x02\x1E\x03\n\x00\x02\x00\x00\x00\x8Cr}S\x9F$j\xC5\x05\x00\x00\x00\x05\x00\x00\x00\a\x00\x18\x00\x00\x00\x00\x00\x01\x00\x00\x00\xA4\x81\x00\x00\x00\x00foo.txtUT\x05\x00\x03\x17D\xA5aux\v\x00\x01\x04\xE8\x03\x00\x00\x04\xE8\x03\x00\x00PK\x01\x02\x1E\x03\n\x00\x02\x00\x00\x00\x90r}S\x90\xFA\x8A\x10\x05\x00\x00\x00\x05\x00\x00\x00\a\x00\x18\x00\x00\x00\x00\x00\x01\x00\x00\x00\xA4\x81F\x00\x00\x00bar.txtUT\x05\x00\x03 D\xA5aux\v\x00\x01\x04\xE8\x03\x00\x00\x04\xE8\x03\x00\x00PK\x05\x06\x00\x00\x00\x00\x02\x00\x02\x00\x9A\x00\x00\x00\x8C\x00\x00\x00\x00\x00"

Zip::File.open_buffer(archive) do |archive|
[...]

PR #576, opened 2024-04-09
Zip::File#write_buffer doesn't assume we're writing to the original IO

casperisfine:
`write_buffer` may be called with another IO than the one that was used to open/read the file. Hence even if there is no change, the caller may expect the archive to be written in the IO.

NB: a bunch of test fails, but I'm not sure I fully understand the API and that my assumptions are correct, hence why I'm opening early.

Perhaps we're using the API incorrectly and we ought to rewrite our code?

### DIFF
--- a/lib/zip/file.rb
+++ b/lib/zip/file.rb
@@ -412,8 +412,6 @@ module Zip
 
     # Write buffer write changes to buffer and return
     def write_buffer(io = ::StringIO.new)
-      return io unless commit_required?
-
       ::Zip::OutputStream.write_buffer(io) do |zos|
         @entry_set.each { |e| e.write_to_zip_output_stream(zos) }
         zos.comment = comment
--- a/test/file_test.rb
+++ b/test/file_test.rb
@@ -139,10 +139,6 @@ class ZipFileTest < MiniTest::Test
 
     zf = ::Zip::File.open_buffer(string_io)
     assert zf.entries.map(&:name).include?('zippedruby1.rb')
-
-    # Ensure the buffer isn't changed.
-    assert_same string_io, zf.write_buffer(string_io)
-    assert_equal(data, string_io.string)
   end
 
   def test_open_buffer_no_op_does_not_change_file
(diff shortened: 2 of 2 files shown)


hainesr:
To be honest this is an especially horrible bit of the API, and not one where I really understand the thinking behind it - it's from way before my time as maintainer.

The issue with doing what you've done in this PR is that it regresses the other fix we put in place to make sure that `write_buffer` doesn't change things when it simply opens a buffer for reading (#280, #512, *et al*).

My gut feeling is that this is a bit hairy to fix properly in a minor release. Probably what needs to happen is that `Zip::File` keeps a hold of the underlying IO so that it doesn't need to be passed in explicitly to `write_buffer`, then it can just return the unchanged (or changed) IO. The other thing we could do is go back to returning `nil` if there's no change, then at least you could test for that: if you get an IO back it has changed/been updated, and if you get `nil` back your original IO is still valid and unchanged.

I would be interested in your thoughts on this.

casperisfine:
> My gut feeling is that this is a bit hairy to fix properly in a minor release.

Right, but it also got broken in a minor release 😛 .

> I would be interested in your thoughts on this.

So we're not huge users of the gem, and I am only the guy that's updating the dependency and has limited knowledge about the code that uses this API.

But out of the top of my head, what you say make sense but seems like a long term refectoring given that right now `File` isn't tied to an IO.

Perhaps a short term solution is either:

  - `write_buffer` takes an extra parameter to always write regardless of changes
  - Add another method (no idea about the name) that essentially work like `write_buffer` used to.

hainesr:
> > My gut feeling is that this is a bit hairy to fix properly in a minor release.
> 
> Right, but it also got broken in a minor release 😛 .

Yes, you're right.

I feel like I really should revert `write_buffer` to it's broken-in-some-ways state and fix the issue people were seeing a different way. I could change `open_buffer` to only write if needed, and still return the IO either way (we have the IO at that point). I think that would fix the issue and preserve the functionality we have now.

PR #577, opened 2024-04-09
Fix deprecation in Entry#get_input_stream

casperisfine:
Ran our app with warnings on, and found this one was coming from RubyZip itself.


hainesr:
Good catch, thank you.

I suspect there may be more of these...

PR #578, opened 2024-04-26
Fix misspell in variable name

Ph0tonic:
Just fix a misspell in a variable name


hainesr:
Thanks!

This has been merged manually.

Issue #573, opened 2024-03-22
AES Encryption ?


laurent-martin:
Any plan to support AES Encryption,
what is described here:
https://www.winzip.com/en/support/aes-encryption/

I see that python has support for it, not Ruby as I can see.

If there is an existing extension, or gem, interested to know, what is the status for AES support in this gem ?

I have seen that there was:
https://github.com/rubyzip/rubyzip/pull/179

But it seems that changes to this gem since that date make it not straightforward to re-integrate.

hainesr:
Hi, AES encryption is intended to be supported in future, but I won't have time to include it in 3.0, sorry.

PR #579, opened 2024-04-26
Adds support AES decryption

jplot:
#573 #179 @Ph0tonic


lib/zip/crypto/aes_encryption.rb
```diff
@@ -0,0 +1,67 @@
...
+      }
+      @counter = 0
+    end
+
+    def header_bytesize
+      @headers[:salt_length] + VERIFIER_LENGTH
+    end
+
+    def gp_flags
+      0
```
Ph0tonic, line 30: As described below, the general purpose flag should be set to 1:
>  As for any encrypted file, bit 0 of the "general purpose bit flags" field must be set to 1 in each AES-encrypted file's local header and central directory entry.
- https://www.winzip.com/en/support/aes-encryption/
```suggestion
      1
```

lib/zip/crypto/aes_encryption.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      while amount_to_read > 0
+        @cipher.iv = [@counter + 1].pack('Vx12')
+        begin_index = BLOCK_SIZE * @counter
+        end_index = BLOCK_SIZE * @counter + [BLOCK_SIZE, amount_to_read].min - 1
+        decrypted_data << @cipher.update(encrypted_data[begin_index..end_index])
+        amount_to_read -= BLOCK_SIZE
+        @counter += 1
+      end
+
```
Ph0tonic, line 49: We are missing the check of the CRC, we should check that the decrypted content hasn't been tampered.

Ph0tonic (commented):
Just a few initial remarks from what I have read.

lib/zip/crypto/aes_encryption.rb
jplot: I've tried to integrate the AES integrity check, but it doesn't work.
If you have an idea, I'd love to hear it.

https://github.com/jplot/rubyzip/commit/21bdb688f2fc11a98d7855b78ace04509a2e49ea

jplot:
Hi @hainesr,

What do you think about merging this first version without integrating the integrity test?
And make in a second time the integrity test with the possibility of creating an AES archive.

jplot:
Hi @hainesr,

ready for review.

lib/zip/crypto/aes_encryption.rb
```diff
@@ -0,0 +1,116 @@
...
+    }.freeze
+
+    def initialize(password, strength)
+      @password = password
+      @strength = strength
+      @bits = BITS[@strength]
+      @key_length = KEY_LENGTHS[@strength]
+      @salt_length = SALT_LENGTHS[@strength]
+    end
+
```
Ph0tonic, line 54: If we had decryption, we should also implement encryption. This would provide a way to test the full stack.

lib/zip/crypto/decrypted_io.rb
```diff
@@ -31,12 +34,24 @@ def buffer
...
-      @decrypter.decrypt(@io.read(CHUNK_SIZE))
+      chunk_size = [CHUNK_SIZE, @compressed_size - pos].min
+      @decrypter.decrypt(@io.read(chunk_size))
+    end
+
+    def check_aes_integrity
+      return unless @decrypter.kind_of?(::Zip::AESDecrypter)
+      return unless input_finished?
+
+      @decrypter.check_integrity(@io.read(::Zip::AESEncryption::AUTHENTICATION_CODE_LENGTH))
```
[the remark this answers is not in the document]
jplot: I'll look into it in the next few days, thx.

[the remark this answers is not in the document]
jplot: @Ph0tonic I reread the spec and the integrity check must be performed in all cases.
The only difference is that with AE-1, you also get a CRC

lib/zip/crypto/aes_encryption.rb
jplot: I'm not going to have time to implement encryption

hainesr:
Right, I've rebased everything and opened a new PR for this: #645.

I'll close this one now as all the code from that is in there.

@jplot and @Ph0tonic, please do have a look if you have time.

Issue #623, opened 2025-01-05
2.4: changelog + release + missing `.0` in tag?


tagliala:
Hello,

Can we have the changelog + github release for v2.4.0?

Also I've noticed that it has been tagged and released as `2.4`, but I was expecting `2.4.0`, like any other `.0` patch release

Ref: https://rubygems.org/gems/rubyzip/versions

![image](https://github.com/user-attachments/assets/a2c19ef4-5230-42be-b432-2789d908a650)

hainesr:
Gah, complete brain-fade from me, sorry.

Unfortunately I now can't push a version `2.4.0` as Rubygems complains that the canonical version number is taken, even if I yank the `2.4` version.

I guess I need to go to 2.4.1 then...

hainesr:
Done. Sorry for the mix-up with 2.4.

PR #625, opened 2025-01-06
Enhance changelog and readme

tagliala:
- Keep the `master` changelog up to date until with 2.4 releases. This should help with dependabot's sourced changelog. Example: https://github.com/alphagov/locations-api/pull/551
- Fix the Ruby versions known working with 2.x series

PR #628, opened 2025-01-06
doc: copy changelog entries for v2.4 and v2.4.1 from the 2.4 branch into `main`

flavorjones:
See #623


tagliala:
Hi, I did the same in #625, but I forgot to add "Close #623" (which is already closed btw)

hainesr:
Hi, thanks for this but I already did it on my local master. Will push it now.

PR #630, opened 2025-01-08
Fix CI against JRuby, JRuby-head, and Windows

tagliala:
Additionally:
- Use `latest` rubygems, instead of specifying a version and keeping
  it up to date
- Bump `rake` dependency to `~> 13.2.0` to allow tests to pass against
  Windows

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,7 +25,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby }}
-        rubygems: '3.5.23'
+        rubygems: latest
         bundler-cache: true
 
     - name: Run the tests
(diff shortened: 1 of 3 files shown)


tagliala:
This is fixing both `jruby` and `jruby-head`. Failures on windows caused by missing ostruct?

Earlopain:
You need to allow newer `rake` to fix that

tagliala:
> You need to allow newer `rake` to fix that

Thanks, I actually did not notice that stable development gems are locked at patch level (and non-stable are locked at minor level).

It looks pretty unusual to me. I wonder why if a new PR should be proposed to relax the dependencies for stable libraries following semver (is rake following semver?  there are no references in the code)

tagliala:
JRuby 9.4.10.0 has been released

However:
- It is still not available on CI: https://github.com/ruby/setup-ruby/pull/697
- I guess it may fail the same, because of the failure on jruby-head

hainesr:
Thank you for your work on this.

I have relaxed the gem dependencies and merged this pull request on top of that, and it looks like it has worked!

(Closing this manually.)

PR #637, opened 2025-07-30
Eagerly require gem version

Earlopain:
v3 is released and contains breaking changes. I was a bit surprised that I had to manually require this file.

Can't be helped now of course but it will help with future major versions (if any).

Nodoc on it also seems wrong. To support both v2 and v3, I have to know which version I'm using.


hainesr:
Apologies, this is definitely an oversight on my part, and thanks for the PR.

I'm actually in the middle of re-working the whole `require` tree across the library because it has become a real mess over the years. I will work this PR in as appropriate.

Earlopain:
Thanks (:

PR #639, opened 2025-08-04
entry: escape filename pattern when matching in `#name_safe?`

yosangwon:
After upgrading to 3.0.0, I found filename with parentheses treated as "not safe".

```
[1] pry(main)> require 'zip/version'
=> true
[2] pry(main)> Zip::VERSION
=> "2.4.1"
[3] pry(main)> Zip::File.open("/Users/syi/projects/sample.zip").entries.first.name
=> "filename with (parentheses)"
[4] pry(main)> Zip::File.open("/Users/syi/projects/sample.zip").entries.first.name_safe?
=> true
```

```
[1] pry(main)> require 'zip/version'
=> true
[2] pry(main)> Zip::VERSION
=> "3.0.0"
[3] pry(main)> Zip::File.open("/Users/syi/projects/sample.zip").entries.first.name
=> "filename with (parentheses)"
[4] pry(main)> Zip::File.open("/Users/syi/projects/sample.zip").entries.first.name_safe?
=> false
```

It looks like regular expressions matching for Windows drive mapping doesn't have any escapes, so I added it.

### DIFF
--- a/lib/zip/entry.rb
+++ b/lib/zip/entry.rb
@@ -230,7 +230,7 @@ module Zip
       return false unless cleanpath.relative?
 
       root = ::File::SEPARATOR
-      naive = ::File.join(root, cleanpath.to_s)
+      naive = Regexp.escape(::File.join(root, cleanpath.to_s))
       # Allow for Windows drive mappings at the root.
       ::File.absolute_path(cleanpath.to_s, root).match?(/([A-Z]:)?#{naive}/i)
     end


hainesr:
Aha yes, we move from a simple compare (which meant we weren't supporting Windows paths properly) to a regex match in 3.0.

I think this is good, thanks, and I'll merge when I've got a suitable clean point to do so 🚀

hainesr:
Merged manually, thanks @yosangwon! 🚀

PR #643, opened 2025-08-19
fixed sysread having a frozen string literal as default

nemesit:
Without this '' will be frozen and cause issues when frozen string literals are enabled # frozen_string_literal: true
And third party gems sometimes do a .sysread without parameters


hainesr:
Hi @nemesit, thanks for catching this.

The linter is complaining that `+''` would be more performant though.

And would you mind adding a test? No worries if you don't have time - I can do it.

Thanks again

PR #661, opened 2026-04-18
Refactor `InputStream` and `AbstractInputStream`.

hainesr:
This pull request significantly refactors and enhances the IO API compatibility and behavior of the `Zip` gem's input stream classes. The main focus is to make the custom IO classes behave more like Ruby's built-in IO, especially regarding encoding handling, buffer management, and line-based reading methods. The changes also add comprehensive documentation and tests for these behaviors.

Key improvements and changes include:

### IO API Compatibility and Encoding Handling

* Refactored the `read` and `sysread` methods across decompressors and input streams to align with Ruby IO conventions: they now properly handle optional output buffers, preserve or force encodings as appropriate (defaulting to `ASCII-8BIT` unless a buffer is provided), and return values consistent with IO semantics. (`lib/zip/ioextras/abstract_input_stream.rb`, `lib/zip/input_stream.rb`, `lib/zip/inflater.rb`, `lib/zip/pass_thru_decompressor.rb`, `lib/zip/crypto/decrypted_io.rb`) [[1]](diffhunk://#diff-0d01947a5bf5d91f390ba8592f7f0d55d152e9b3bfeb6d3c08dee9a337053ac7L6-R207) [[2]](diffhunk://#diff-24aba904eb61cedd0754c81bb6cc17713dc1cf871ef8a010bf06b07fe361a5ddL92-R113) [[3]](diffhunk://#diff-5c44c47f705ba8c5d992ab914554972740440428aa77f573eaa79006f00c4285L8-R21) [[4]](diffhunk://#diff-4be0a4c92c7e684634b4d8a39cdf42d448ed0efe3930c7f045580b45f1e8cb20L10-R18) [[5]](diffhunk://#diff-e78aa814d0f04b496615b891a24052504f6310ae22f37537a7184610782f3f8fL11-R25)

[...]

PR #662, opened 2026-04-20
Fix AbstractInputStream#gets to not assume `a_sep_string` will appear…

ListlessPanda:
The `AbstractInputStream#gets` has logic to check if the number of bytes within the `@output_buffer` are over the limit then sets `over_limit`. It seems that if a separator was passed in along with the number of bytes the code assumes the separator will appear within the bytes read. 

This was picked up when streaming a CSV out of a ZIP where the CSV passes the `a_sep_string` as `\n` and the number of bytes as `8192` and the file contains a single column value that is longer than `8192` bytes

I made a change that if the separator wasn't found and we are above the limit it will return up to the number of bytes and increase the `@pos` by that amount. 

(This is my first pull request please let me know if I need to do anything else)

### DIFF
--- a/lib/zip/ioextras/abstract_input_stream.rb
+++ b/lib/zip/ioextras/abstract_input_stream.rb
@@ -82,9 +82,14 @@ module Zip
           @output_buffer << produce_input
           over_limit = (number_of_bytes && @output_buffer.bytesize >= number_of_bytes)
         end
-        sep_index = [match_index + a_sep_string.bytesize, number_of_bytes || @output_buffer.bytesize].min
-        @pos += sep_index
-        @output_buffer.slice!(0...sep_index)
+        if over_limit && match_index.nil?
+          @pos += number_of_bytes
+          @output_buffer.slice!(0...number_of_bytes)
+        else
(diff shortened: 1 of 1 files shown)


hainesr:
Hi @ListlessPanda,

Thanks for looking at this. I've been doing a load of updates to `AbstractInputStream` recently in #661. Please could you test your error case as you describe above against the code in #661? It may be fixed in there already and I plan to merge and release that soon.

hainesr:
Looking at the tests I added in #661, I think I test the case you mention above, here:
https://github.com/hainesr/rubyzip/blob/af08a4754c6565fd92dff1d3356d5fe974aa6ee3/test/ioextras/abstract_input_stream_test.rb#L360-L371

You can see that some of the returned lines are truncated by the limit before a separator is found. Is this what you were fixing?

hainesr:
I've now merged #661 and released as v3.3.0, so please can you test that and see if it fixes the issue you're seeing?

hainesr:
I think this has been fixed in v3.3, so closing this PR, but please reopen if not the case and I will have another look.

PR #665, opened 2026-06-26
Fixed decrypted_io to only perform integrity once

ListlessPanda:
If read is called on an entry where the Zip::DecryptedIo::CHUNK_SIZE will reach the end (thus input_finished?) but maxlen limits the output we will still do the integrity check on any follow up read which will then cause the integrity check to read 10 bytes further and cause an integrity error. 

I added a change so that if the integrity check has already been done (meaning we have reached the end) it won't run again on buffer reads,


hainesr:
Hi @ListlessPanda,

Thanks for this. I'll have a look in more detail over the next couple of days.

Don't worry about the failing linter; it's just a bit of trailing whitespace which I can fix easily enough.

hainesr:
Hi @ListlessPanda,

Thanks, I've now merged this minus the changes related to returning empty strings. It may well be that your changes were correct, but they may also have caused backwards-incompatibility at this stage. I will fix all these issues in the next major version.

Will release this code soon, thanks again!

PR #667, opened 2026-06-28
Support passing decrypter to encrypted entry via get_input_stream

ListlessPanda:
This adds the ability for someone to pass a decrypter to an encrypted entry easily.

If someone were to grab a Zip::Entry from the CentralDirectory this will still allow them to easy pass a decrypter to the created InputStream for easy adhoc decryption. 

Currently if a zip is encrypted we can only access the decrypted entries via the InputStream and it doesn't not integrate with the CentralDirectory meaning we would need to read through the entries sequentially. With this we can have a bit of both worlds. Allow random entry access and decrypt them if they are encrypted.

I was looking into potentially passing the decrypter to a Zip::File and then it passes it to the CentralDirectory that will initialize entries with a `@decrypter` and then they would use that for the InputStream but this be a larger rework and might be completely unnecessary.

### DIFF
--- a/lib/zip/entry.rb
+++ b/lib/zip/entry.rb
@@ -638,8 +638,9 @@ module Zip
     end
 
     # Returns an IO like object for the given ZipEntry.
+    # Optional decrypter can be provided for an encrypted entry
     # Warning: may behave weird with symlinks.
-    def get_input_stream(&block)
+    def get_input_stream(decrypter: nil, &block)
       if ftype == :directory
         yield ::Zip::NullInputStream if block
         ::Zip::NullInputStream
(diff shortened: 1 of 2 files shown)


hainesr:
Hi @ListlessPanda,

Thanks for this. I actually have this on a dev branch already, but I had got bogged down trying to also add it to the `Zip::File` API as well. I think you're right that we might as well just do `Zip::Entry` first and then consider if going any further is necessary later. Something to note on that score is that there is nothing in the spec to say that the same encryption (or indeed password) must be used for every file in the Zip, so making this a per-`Entry` operation does make sense.

Your code is exactly the same as mine, but actually tested, so I'll merge this and release it in 3.5 at some point soon-ish.

Thanks so much!

PR #668, opened 2026-06-28
Support setting the encoding of the input stream

### LINKED ISSUE #546: IOExtras::AbstractOutputStream doesn't implement `external_encoding`
With nokogiri 1.14.0 and rubyzip 2.3.2, running the code below fails with `NoMethodError: undefined method 'external_encoding' for #<Zip::OutputStream:0x00000001241ec308>`:

```ruby
doc = Nokogiri::XML("<test/>")
Zip::OutputStream.open('test.zip') do |out|
  out.put_next_entry('test.xml')
  doc.write_to(out, encoding: 'UTF-8')
end
```

I'm not sure why Nokogiri calls `external_encoding` as as from the [IO](https://ruby-doc.org/core-3.1.2/IO.html#method-i-external_encoding) and [Encoding](https://ruby-doc.org/core-3.1.2/Encoding.html#top) docs `external_encoding` appears to be only for reading, but it also seems from the docs like just returning `nil` would be a legit implementation.

(Cf. https://github.com/sparklemotion/nokogiri/issues/2773)

ListlessPanda:
Currently the `AbstractInputStream#read` supports passing an `out_string` to change the encoding of the output. 
But in some cases we are not the ones calling the `read` or `gets` and these are being done inside some parser like `CSV`. In most cases the parser doesn't `force_encode` the output of the input stream and tries to parse on the wrong encoding. 

With this change we can set the encoding when we get the input stream for the zip. So that the data being sent to the parsers will be in the expected encoding.

I tried following a similar pattern I had seen in `StringIO` and `IO` where you can pass the encoding along in the initialize or call `set_encoding` on the stream to change it.

### DIFF
--- a/lib/zip/input_stream.rb
+++ b/lib/zip/input_stream.rb
@@ -51,8 +51,9 @@ module Zip
     #
     # @param context [String||IO||StringIO] file path or IO/StringIO object
     # @param offset [Integer] offset in the IO/StringIO
-    def initialize(context, offset: 0, decrypter: nil)
-      super()
+    # @param encoding [Encoding] encoding of the decompressed stream
+    def initialize(context, offset: 0, decrypter: nil, encoding: Encoding::ASCII_8BIT)
+      super(encoding: encoding)
       @archive_io = get_io(context, offset)
       @decompressor = ::Zip::NullDecompressor
       @decrypter = decrypter
@@ -107,7 +108,7 @@ module Zip
       output = produce_input(maxlen)
 
       if out_string.nil?
-        output.force_encoding(Encoding::ASCII_8BIT)
+        output.force_encoding(@encoding)
       else
         encoding = out_string.encoding
         out_string.replace(output).force_encoding(encoding)
--- a/lib/zip/input_stream.rb
+++ b/lib/zip/input_stream.rb
@@ -51,9 +51,10 @@ module Zip
     #
     # @param context [String||IO||StringIO] file path or IO/StringIO object
     # @param offset [Integer] offset in the IO/StringIO
-    # @param encoding [Encoding] encoding of the decompressed stream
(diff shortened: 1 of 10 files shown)


hainesr:
Hi @ListlessPanda,

Thanks for this. I'm interested in this and think it'll help generally, as you note. I have a couple of questions/notes:
* There has been an issue submitted (#546) requesting this behaviour on output streams as well. Is it worth moving this code up to perhaps the `FakeIO` class so it can be applied to both input and output streams? (Even if for output streams it's mostly a no-op.)
* It looks like the options to [`IO::new`](https://rubyapi.org/3.4/o/io#class-io-open-options) require either `external_encoding:` and `internal_encoding:` to be set separately, or to specify `encoding: <external:internal>` and provide both at once. In the spirit of mimicking `IO` as closely as possible, perhaps we should follow this convention too.
* Then we would also need methods to get/set `internal_encoding` and `external_encoding`, and `set_encoding` should mirror the `<external:internal>` pattern above.

Does that make sense? Happy to work on this as well.

hainesr:
Hi @ListlessPanda,

I've found time to look at this again and have (force) pushed some updated functionality to this PR. It basically works like `IO` as I talked about above, with `external_encoding` and `internal_encoding` set internally in `FakeIO`.

Edit (because I got internal and external coding mixed up last night, as per [this example](https://ruby-doc.org/core-3.1.1/Encoding.html#class-Encoding-label-IO+encoding+example)):
I think `external_encoding` (what is in the actual files) is always `BINARY` for Zips, so maybe we should just set it to `BINARY` and ignore user attempts to change it? I need to update my recent changes to use `internal_encoding` for the output of the input streams, which I'll do today and re-push.

Any thoughts on the updates welcome. Thanks again for kicking this work off.

hainesr:
I have now fixed up my confusion between internal and external encoding and also wired this into output streams as well.

PR #669, opened 2026-06-30
Clarify README on storing non-English filenames as UTF-8

avaice:
Thanks for this useful library!

Fixed the "Non-ASCII Names" section, where the Windows version range was reversed.

`Zip.unicode_names = true` sets the ZIP EFS flag (bit 11), marking filenames as UTF-8. The Windows built-in ZIP handler honors this flag only on Windows 7 or later; earlier versions ignore it. The previous wording (`Windows(pre 7)`) was the opposite of the actual behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,7 +367,7 @@ Zip.continue_on_exists_proc = true
 
 ### Non-ASCII Names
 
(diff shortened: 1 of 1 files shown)


hainesr:
Thank you for this clarification!

PR #670, opened 2026-08-10
Fix the link to Ruby doc in REAMD to the latest version

y-yagi:
There is a link to Ruby doc in README, but that points to the old version(2.2).
This commit fixes the link to the latest and official doc.


hainesr:
Thanks for this!

Merged manually...

PR #672, opened 2026-08-28
Load openssl only when AES encryption is used

tas50:
## Problem

`lib/zip/crypto/aes_encryption.rb` requires `openssl` at the top of the file, and `zip.rb` requires that file unconditionally. `openssl` in turn pulls in `ipaddr` and `socket`, so **every** `require 'zip'` pays for the whole TLS stack — even though only AES-encrypted archives need it.

It is by a wide margin the single largest cost of loading rubyzip:

```
   53.00  zip
    7.71    tempfile
   23.02    zip/crypto/aes_encryption      <-- 43% of total
   22.72      openssl
    5.63        openssl.so
    5.08        ipaddr
    3.44          socket
```

## Fix

Every `OpenSSL` reference in the file sits inside `AESDecrypter#reset!` (`OpenSSL::KDF.pbkdf2_hmac`, `OpenSSL::Cipher::AES`, `OpenSSL::HMAC`, `OpenSSL::Digest`). The constants other files read — `AESEncryption::VERSIONS` from `entry.rb` and `AESEncryption::AUTHENTICATION_CODE_LENGTH` from `input_stream.rb` — are plain Ruby and need nothing loaded.

Moving the require into `AESEncryption#initialize` covers every path: `AESDecrypter` is the only includer of the module, and `reset!` cannot be reached without constructing one first.

## Measurements

Ruby 4.0.6 (arm64-darwin), best of seven runs:

| | `require 'zip'` | files loaded |
|---|---:|---:|
| before | 45.69 ms | 75 |
| after | 24.67 ms | 54 |
| | **−21.02 ms (46%)** | **−21** |

Confirmed directly:

```ruby
require 'zip'
# openssl loaded? => false
Zip::AESDecrypter.new('pw', Zip::AESEncryption::STRENGTH_256_BIT)
# openssl loaded? => true
```

## Tests

`bundle exec rake test`: **415 runs, 0 failures** (was 412 + the 3 added here).

The existing AES coverage already exercises the deferred path end to end — `test/crypto/aes_encryption_test.rb` constructs decrypters and runs `reset!` + `decrypt`, and `test/encryption_test.rb` reads the real `zip-aes-128` / `zip-aes-256` fixtures. Both pass unchanged.

The three added tests assert in a subprocess (the suite itself has already loaded openssl via the AES tests) that:
1. requiring `zip` does not load openssl — **fails against the previous code**
2. constructing an `AESDecrypter` does load it
3. AES decryption still produces the correct plaintext with no eager require

`rubocop` clean on both changed files.

### DIFF
--- a/lib/zip/crypto/aes_encryption.rb
+++ b/lib/zip/crypto/aes_encryption.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'openssl'
-
 module Zip
   module AESEncryption # :nodoc:
     VERIFIER_LENGTH = 2
@@ -45,6 +43,11 @@ module Zip
     }.freeze
 
     def initialize(password, strength)
+      # Loaded here rather than at the top of the file so that `require 'zip'`
+      # does not pull in openssl. Only AES-encrypted archives need it, and it
+      # is the single largest cost of loading this library.
+      require 'openssl'
+
       @password = password
       @strength = strength
       @bits = BITS[@strength]
--- a/test/require_test.rb
+++ b/test/require_test.rb
@@ -6,9 +6,17 @@ class RequireTest < Minitest::Test
   # Guards the deferred `require 'openssl'` in Zip::AESEncryption#initialize.
   # These run in a subprocess because the test suite has already loaded
   # openssl by way of the AES tests.
+
+  # Interpreter options set in the environment are inherited by the
+  # subprocess, where they can both add noise to its output and preload
+  # libraries -- either of which invalidates what is measured here. Our own
+  # CI runs the suite with `RUBYOPT=-v`, which makes every subprocess print
+  # its version banner ahead of the script's own output.
+  CLEAN_ENV = { 'RUBYOPT' => nil }.freeze
+
   def subprocess(script)
     lib = File.expand_path('../lib', __dir__)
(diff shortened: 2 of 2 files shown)


hainesr:
Hi @tas50,

Thank you very much for this! It's a good catch, especially as AES decryption really isn't used all that much in real world usage of the library (I think...).

The change and tests all look obviously good to me, although they are failing when run through Actions. After some digging about it looks like the `-v` switch is being inserted somewhere - perhaps GitHub have aliased `ruby` to be `ruby -v` in their Actions runners - so those tests always output something like this:

```ruby
"ruby 3.2.0 (2022-12-25 revision a528908271) [arm64-darwin21]\ntrue"
```

I guess we can fix this by testing the subprocess output with a regex instead? Then we'd have tests that pass locally (without the `-v`) and within Actions. I had a quick look to see if there's a `ruby` CLI switch to reverse the `-v` that we could add into our tests, but nothing jumped out at me.

If you'd like to have a look at fixing this, that would be great, but equally I am very happy to take it from here. This PR is a great start.

Many thanks again!

tas50:
@hainesr See if that fixes it 🤞

hainesr:
Ah ha, thanks. I had forgotten we add `-v` to the `RUBYOPT` environment in Actions.

Thanks for this additional fix; it looks like it's all passing now, and it is better to not use a regex in the tests...

PR #674, opened 2026-08-30
Forward options from blockless OutputStream.open

OskarEichler:
## Summary

Forward encrypter: and suppress_extra_fields: through the blockless Zip::OutputStream.open path instead of silently discarding them.

## Reproduction

The current blockless path returns new(file_name) while the block path forwards both public keywords. A dual-Ruby external model observes the missing encrypter and extra-field suppression on current main; this candidate preserves them.

## Verification

- Added an upstream regression test covering both forwarded options
- Focused output stream tests: 28 runs, 88 assertions, 0 failures
- Ruby 4.0.6 and 3.2.11 full suite before the follow-up: 416 tests, 2,857 assertions, 0 failures

### DIFF
--- a/lib/zip/output_stream.rb
+++ b/lib/zip/output_stream.rb
@@ -53,7 +53,13 @@ module Zip
       # stream is passed to the block and closed when the block
       # returns.
       def open(file_name, encrypter: nil, suppress_extra_fields: false)
-        return new(file_name) unless block_given?
+        unless block_given?
+          return new(
+            file_name,
+            encrypter:             encrypter,
+            suppress_extra_fields: suppress_extra_fields
+          )
+        end
 
         zos = new(file_name, stream: false, encrypter: encrypter,
                   suppress_extra_fields: suppress_extra_fields)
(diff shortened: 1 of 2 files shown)


hainesr:
Hi @OskarEichler,

Many thanks for this (and the other PRs you submitted as well)!

This one looks obviously good, and corrects what I can only assume is an oversight from years ago.

One question: I don't understand what you by
> No repository tests were changed because the consuming audit prohibits upstream test edits.

Although GitHub Actions aren't run by default for new contributors (I have approved them all now), adding to the tests in PRs is fine/encouraged. Maybe I have mis-understood?

Thanks again.

OskarEichler:
You understood correctly; that sentence came from an internal audit constraint that I mistakenly carried into this upstream PR. It does not belong here. I removed it from the description and added a regression test that exercises the blockless path with both encrypter and suppress_extra_fields. The focused file now passes with 28 runs, 88 assertions, and no failures; targeted RuboCop is also clean.

PR #676, opened 2026-08-30
Retain required Zip64 offset fields

OskarEichler:
## Summary

Keep the Zip64 extra field whenever either the original size or relative header offset requires 64-bit storage.

## Reproduction

`Zip64#suppress?` currently combines the size and offset conditions as `!large_size || large_offset`. This suppresses Zip64 when only the offset is large and also suppresses it when both are large. A focused dual-Ruby external model covers size-only, offset-only, both-large and neither-large cases; current `main` fails the required offset cases and this candidate passes all four.

## Verification

- Ruby 4.0.6 and 3.2.11: 416 tests, 2,857 assertions, 0 failures

No repository tests were changed because the consuming audit prohibits upstream test edits.

### DIFF
--- a/lib/zip/extra_field/zip64.rb
+++ b/lib/zip/extra_field/zip64.rb
@@ -59,8 +59,8 @@ module Zip
     # We can suppress the zip64 extra field unless we know the size is large or
     # the relative header offset is large (for central directory entries).
     def suppress?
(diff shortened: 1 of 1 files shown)


hainesr:
Erk, this one was definitely on me. Good catch, thanks.

PR #677, opened 2026-08-30
Parse legacy Unix owner identifiers

OskarEichler:
## Summary

Assign UID and GID values parsed from the legacy Unix extra field instead of retaining the constructor's initialized zero values.

## Reproduction

`OldUnix#initialize` sets `@uid` and `@gid` to zero, then `#merge` uses `||=` for the parsed values. Because zero is truthy, nonzero identifiers in a valid legacy field are never exposed. A dual-Ruby external binary-field model reproduces the zero results on current `main` and reads the encoded UID/GID with this candidate.

## Verification

- Ruby 4.0.6 and 3.2.11: 416 tests, 2,857 assertions, 0 failures

No repository tests were changed because the consuming audit prohibits upstream test edits.

### DIFF
--- a/lib/zip/extra_field/old_unix.rb
+++ b/lib/zip/extra_field/old_unix.rb
@@ -24,8 +24,8 @@ module Zip
       return if !size || size == 0
 
       atime, mtime, uid, gid = content.unpack('VVvv')
-      @uid ||= uid
-      @gid ||= gid
+      @uid = uid unless uid.nil?
+      @gid = gid unless gid.nil?
       @atime ||= atime
       @mtime ||= mtime # rubocop:disable Naming/MemoizedInstanceVariableName
     end


hainesr:
Thanks for this PR as well.

I don't think this extra field is used anywhere in the library, so it's just there as a known field that is parsed and, I guess, rewritten if the archive is then saved again. Some users of the library might need this extra field, of course, but it's not used internally like universal time is, for example.

You've definitely found a bug - given that zero is truthy in Ruby it won't store the actual `uid` or `gid` internally. Even worse, it will set them both to zero - which is root! Your fix potentially preserves this behaviour by leaving `uid` and `gid` at zero if there are any issues reading from the extra field, so I think I will fix this by setting everything to `nil` in the constructor instead.

Thanks for the catch!

PR #680, opened 2026-09-15
Fix `AbstractInputStream#gets` to not split multi-byte characters in gets

ListlessPanda:
Currently `AbstractInputStream#gets` applies the `limit` as a plain byte count, so if the limit lands in the middle of a multi-byte character the line comes back with that character cut in half. Which should happen with `#read` but `#gets` is suppose to extend past the hard limit to prevent cutting characters (as noted in https://docs.ruby-lang.org/en/3.4/IO.html "(A multi-byte character will not be split, and so a returned line may be slightly longer than the limit)")

I made a change so that the limit is worked out first and then the cut is moved up to the end of the character it landed in. Ruby does this in C with `rb_enc_right_char_head` which isn't exposed to Ruby, so rather than reimplementing it I let a `StringIO` over the bytes around the cut work out the boundary. It only kicks in when an internal encoding has been set, binary reads still cut exactly on the byte like before.

I also added tests for the different limit forms along with UTF-16LE and a 4 byte character, and one that reads a multi-byte entry out of an encrypted zip since the data arrives in chunks there and a character can straddle two of them.

I did test this against a bunch of files containing random characters and random limits to find any other edge cases. But they did match other `IO` like `gets`. 

This was found when I was reading a CSV entry out of a ZIP. 
In the CSV gem it has a method `CSV::Parser#resolve_row_separator` that when the `separator` is set to `:auto` it will do `break unless sample = @input.gets(nil, chunk_size)` and if `chunk_size` happen to cut exactly on a character the parser will raise `InvalidEncoding` because the `valid_encoding?` goes `false` on the `sample` later on.


ListlessPanda:
So this seems to have also found that exact case being broken in [truffleruby](https://github.com/truffleruby/truffleruby) 

I've gone ahead and made a [PR](https://github.com/truffleruby/truffleruby/pull/4468) on their repo with a fix as well.

hainesr:
Hi @ListlessPanda,

Thanks for this. It's late here so I will look at this properly over the next couple of days, but it looks sensible after a first pass.

I'd be nervous about breaking rubyzip for Truffleruby so I'll keep my fingers crossed that truffleruby/truffleruby#4468 is looked at/merged soon!

eregon:
One quick thought, what about using a Regexp (`/./`) to detect the end of character in a portable way and without needing StringIO?

PR #681, opened 2026-09-19
Bound Inflater memory use on highly compressible entries.

ThomasSevestre:
Inflater fed Zlib 32 KiB of compressed data per step. Deflate expands up to ~1000:1, so one step could add over 30 MiB to the internal buffer, which was then drained in small slices, each moving the whole remainder.

Feed Zlib 4 KiB per step instead. The read loop already stops once the buffer holds the requested length, so the excess inflated per read is now bounded by ~4 MiB. Output is unchanged, as are Decompressor::CHUNK_SIZE and the BufError retry.

20 MB of repeated XML (71 KB compressed) read in 16 KB chunks:

  before: 3.1 s, peak buffer 9.4 MB
  after:  1.0 s, peak buffer 1.2 MB

Ordinary files (25 MB zip of 749 real files) read whole, in chunks or extracted are as fast as before. Whole reads of large, barely compressible entries are up to ~5% slower on Ruby 3.4 (unchanged on 4.0).


hainesr:
Hi @ThomasSevestre,

Thanks for this. It seems like a good precaution to limit how much is inflated in one go.

I have removed the conflict in the `Changelog.md` (unlucky timing that I cut a release last night!) for now. I will populate `Changelog.md` at the time of release.

Thanks!
Rob

hainesr:
A further thought after looking at this more closely: perhaps we should make the input chunk size configurable? This would allow users to tailor the value (and therefore some level of performance) to their expected inputs.

What do you think? And if configurable, what should the default be?

Thanks again.

ThomasSevestre:
Hello @hainesr,

Indeed I believe it is a good idea to make it configurable, I thought about it after pushing the PR as well!

About the default value, I see 2 solutions:
- use Decompressor::CHUNK_SIZE to keep the current behavior as the default
- use 4096 as proposed in this PR to limit the risk of high memory usage when dealing with unknown input

I'd choose 4096 since it limits memory usage and has arguably better performance.

Let me know your opinion and I'll implement it

Thanks

hainesr:
Amazing, thank you.

I'm happy with 4096 for the default if you think this strikes a good balance?

What should we call the config option? Something like `inflater_chunk_size`? I guess this can sit in the top-level `Zip` namespace with the other config for now as it feels fundamental enough, even though it's specifically for `Inflater`.

I guess we probably need to test a couple of different settings as well. And for the README please could you add a couple of lines in the configuration section on why anyone might want to change the default for different expected use cases?

Thanks for your help with this!

ThomasSevestre:
Hello @hainesr,

I pushed the changes with the default configuration set to 4096.

I tested against a few zip files, comparing to the previous 32k setting:

On incompressible files (zip of images, firmware, etc.): the new default value increases processing time by 11%
On highly compressible files (XML, CSV, etc.): the new default value improves processing time by 50%

Assuming zip is mostly used on compressible files, most people should have improvements on memory usage and CPU usage.

Thanks

